PR #1042, opened 2026-03-29
fix: reduce equity Monte Carlo simulations to prevent CPU timeout

yuta-yoshinaga:
## Summary
- Holdem/Omaha/ShortDeck games were returning 503 on Cloudflare Workers (error 1102 = CPU time exceeded) and timing out on Render free tier due to **50,000 Monte Carlo equity simulations** per API response
- Split simulation count constants into build-tagged files:
  - Native (`!js`): 2,000 simulations (~1s on free tier servers)
  - WASM (`js && wasm`): 200 simulations (fits within Workers CPU limits)
- Fix biome import ordering in MemoryPage.tsx

## Test plan
- [x] All Go tests pass (`go test -tags test ./...`)
- [x] Native and WASM builds compile
- [x] `golangci-lint run ./...` clean
- [x] `bun run check` passes
- [ ] Verify holdem/omaha/shortdeck return 200 on Cloudflare after deploy
- [ ] Verify holdem/omaha/shortdeck return 200 on Render after deploy


yuta-yoshinaga:
@claude

PR #1334, opened 2026-04-13
fix: run CPU exchanges after betting round transitions to exchange phase

yuta-yoshinaga:
## Summary

- When the human player is the dealer in 5-Card Draw Poker, completing the first betting round transitions to the Exchange phase, but CPU exchange logic was never invoked — leaving `currentTurn` on a CPU player and blocking the human with "It is not your turn"
- Added a post-transition check in `PlayerAction`: if phase moved to Exchange, run `runCpuExchanges()` and auto-advance to second betting if all exchanges are complete
- Added regression test `TestPoker_PlayerAction_TransitionsToExchangeAndRunsCpuExchanges` covering the dealer-is-human scenario

Closes: N/A (bug found during manual play testing)

## Test plan

- [x] All existing Go tests pass (`go test -tags test ./internal/...`)
- [x] New regression test passes
- [x] `golangci-lint run ./internal/domain/` — 0 issues
- [x] Manual CUI test: `echo -e "c\ne 0 1\nq" | go run ./cmd/trumpcards poker` — exchange now works when human is dealer
- [ ] Verify via Web GUI that poker card exchange works in all dealer positions

### DIFF
--- a/internal/domain/Poker.go
+++ b/internal/domain/Poker.go
@@ -271,6 +271,14 @@ func (p *Poker) PlayerAction(action, amount, humanPlayMs int) error {
 
 	p.advanceTurn()
 	p.runCpuActions()
+
+	// ベッティングラウンド完了で交換フェーズに移行した場合、CPU交換を実行
+	if p.round.phase == PokerPhaseExchange && !p.round.gameEndFlag {
+		p.runCpuExchanges()
+		if p.isExchangeComplete() {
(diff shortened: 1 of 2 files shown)


yuta-yoshinaga:
@claude

yuta-yoshinaga:
@claude

yuta-yoshinaga:
@claude

PR #1574, opened 2026-04-28
refactor(frontend): factor solitaire move and simple bet APIs in gameApi (slice 1)

yuta-yoshinaga:
## Summary

`gameApi.ts` had grown to **1476 lines** partly because nine solitaire-style move endpoints and the two single-amount bet endpoints each spelled out the same `exec` wrapper by hand. Adding a new game meant copying 10-line blocks; renaming a payload key meant editing 11 places.

This PR pulls each shape into a tiny factory and reduces every call site to a single declaration, mirroring the existing `createBidPlayApi` / `createVideoPokerApi` pattern:

| Factory | Body | Adopters |
|---|---|---|
| `createSolitaireMoveApi<T, Zone>` | `{ command, from, to, n }` | Canfield, FreeCell, Yukon, Scorpion, Accordion, FortyThieves, Calculation |
| `createSolitaireMoveApiWithConfig<T, Zone, C>` | `{ command, from, to, config, n }` | Klondike, Spider |
| `createBetAmountApi<T, Cmd>` | `{ command, amount }` | LetItRide, RedDog |

Each factory takes a `Cmd` type parameter (defaulting to a broad `SolitaireMoveCommand` union covering `reset` / `move` / `giveup` / `hint` / `autocomplete` / `log` / `undo` / `undo_n` / `draw` / `deal`) so each call site can lock down its own verb set without copy-pasting union literals.

## Why slice 1 only

The issue itself says **「PR を分割」** (split into PRs). This is slice 1, focused on the highest-value low-risk consolidations: the solitaire moves (uniform shape across 9 endpoints) plus the simplest bet APIs. Future slices can extend the pattern to:

- Poker-bet APIs (`pokerApi`, `badugiApi`, `indianpokerApi`) — share `(command, amount, config, humanPlayMs, profile)`
- Single-card-action APIs (`pageoneApi`, `crazyeightsApi`, `ginrummyApi`, `cribbageApi`) — share `(command, cardIndex, config, ...)`
- More exotic shapes (`cassinoApi`, `nertzApi`, `skatApi`, `shitheadApi`) which already use opts-objects

Splitting keeps each PR within reviewable size and keeps `gameApi.test.ts` (3154 lines, 195 cases) churn small per PR.

## Test plan

- [x] `bun run test -- --run gameApi` — 195/195 pass
- [x] `bun run check` — clean (4 pre-existing warnings unrelated to this PR)
- [x] `bunx tsc --noEmit` — clean
- [x] `gameApi.ts` shrinks 1476 → 1421 lines (-55) without changing the wire shape

The factories produce identical request bodies (same keys in the same order, undefined values dropped by `JSON.stringify` exactly as before), so the existing `expect(mockFetch).toHaveBeenCalledWith(...)` assertions in `gameApi.test.ts` keep passing without modification.

Refs #1550

### DIFF
--- a/frontend/src/api/gameApi.test.ts
+++ b/frontend/src/api/gameApi.test.ts
@@ -3,6 +3,7 @@ import {
   actionLogApi,
   baccaratApi,
   blackjackApi,
+  canfieldApi,
   crazyeightsApi,
   daifugoApi,
   doubtApi,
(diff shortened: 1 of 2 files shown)


yuta-yoshinaga:
@claude

PR #1640, opened 2026-05-03
fix: address staging E2E findings (panic recovery + SPA deep links + i18n)

yuta-yoshinaga:
## Summary
Fix four bugs found via Playwright exploratory E2E against `https://go-trumpcards-staging.pages.dev`:

| # | Severity | Bug | Fix |
|---|---|---|---|
| 4 | 🔴 Critical | 11 classic games (sevens, ohhell, oldmaid, durak, daifugo, pageone, whist, president, skat, shithead, tonk) panic on the 2nd Worker request after restoring state from KV. Cloudflare wraps the panic in a 1101 HTML error page (no CORS headers), which the frontend surfaces as `通信エラーが発生しました` with no detail. | **Defensive**: wrap each Worker entry (casino/classic/solo) with a panic-recovery middleware that turns the panic into a JSON 500 with CORS preserved + structured `slog.Error` log. The actual panic message is now in the response body, so we can root-cause via deployed staging. |
| 1 | 🔴 Critical | Path-style deep links like `/hearts` silently render Blackjack — `HashRouter` only sees the hash, and the SPA fallback returns `index.html` which falls through to the default `/` route. | Normalise `/foo` to `/#/foo` in the inline boot script before React mounts. |
| 2 | 🟡 Medium | Unknown hash routes (`/#/notagame`) leave `<main>` empty (no fallback `<Route>`). | Add `<Route path="*" element={<Navigate to="/" replace />} />`. |
| 3 | 🟢 Minor | `gofish` was the only game whose JA title rendered as English ("Go Fish"). | Update `nav.gofish` in `ja/common.json` and per-page `title`/`setup.title` in `ja/gofish.json`. |

## Why panic-recovery instead of root-cause for #4
- Native `go test ./internal/usecase/...` snapshot/restore tests pass for all 11 affected games — the panic only manifests in TinyGo + Cloudflare Worker.
- A minimal TinyGo + WASI repro of the same `[5]uint16`/`[]*Card`/custom `MarshalJSON` patterns runs cleanly, so the trigger is more game-specific.
- Local TinyGo build of the full classic worker OOMs in this dev environment, blocking direct repro.
- The defensive middleware is independently valuable (no more 1101 HTML pages), and once deployed, the JSON 500 body will reveal the actual panic message — turning a 30-min reverse-engineering session into a single staging request. Root-cause fix will follow in a separate PR.

## Test plan
- [x] `go test -tags test ./...` (all green)
- [x] `golangci-lint run ./...` (0 issues)
- [x] `goimports -l` on changed Go files (clean)
- [x] `cd frontend && bun run check` (Biome clean)
- [x] `cd frontend && bun run build` (succeeds)
- [x] `cd frontend && bun run test --run src/pages/GoFishPage` (23 passed)
- [ ] After deploy: confirm a sevens 2nd request now returns a JSON 500 with the panic message (instead of `cf-error-details` HTML). Use the body to identify the root cause and open follow-up PR.
- [ ] After deploy: visit `https://go-trumpcards-staging.pages.dev/hearts` → should land on Hearts (not BlackJack).
- [ ] After deploy: visit `https://go-trumpcards-staging.pages.dev/#/notagame` → should redirect to home.
- [ ] After deploy: visit gofish in JA locale → page heading & tab title should read "ゴーフィッシュ".


yuta-yoshinaga:
@claude

PR #1646, opened 2026-05-04
chore: revert temporary diagnostics from sevens staging bug investigation

yuta-yoshinaga:
## Summary

Cleans up two temporary diagnostic measures left over from the staging Sevens bug investigation, now that root causes are identified and fixed.

| Source | What was added | Why removable now |
|---|---|---|
| #1641 | `slog.Error(\"KV GetString failed\", ...)` branch in `KVSessionProvider.Acquire` | PR was tagged \"will be reverted once the source of the '<' payload is identified.\" Root cause = JS-null-as-\"<null>\"-string from syumai/workers binding, fixed permanently in #1642. The `data_len`/`data_prefix` dumps were already removed by #1642's restructure; this removes the last residual log branch. |
| #1643 | 200-iter cap in `runCpuTurns` | Cap was defensive against a hypothetical hung CPU loop, but the actual bug turned out to be a TinyGo stack overflow (fixed in #1645 by removing slog traces and bumping stack to 128KB). No infinite loop has ever been observed — the cap is speculative defense over the original unbounded loop. |

**Kept** (these are the actual permanent fixes from this investigation):

- `recoverymw` — panic → JSON 500 with CORS headers preserved
- `kvNullSentinel` constant + handling — the real KV binding workaround (#1642)
- `tinygo -stack-size=128KB` in Makefile — necessary for Sevens Reset to succeed

## Test plan

- [x] `go test -tags test ./...` (all green)
- [x] `golangci-lint run ./...` (0 issues)
- [x] `goimports -w` on modified files
- [ ] After merge: confirm staging Sevens reset still returns full JSON
- [ ] After merge: spot-check Sevens Reset/Play/Joker flows in browser

### DIFF
--- a/internal/adapter/controller/kv_session_provider.go
+++ b/internal/adapter/controller/kv_session_provider.go
@@ -69,12 +69,7 @@ func (p *KVSessionProvider[T]) Acquire(id string, factory func() T) (T, SessionR
 
 	if p.ns != nil {
 		data, err := p.ns.GetString(key, nil)
-		switch {
-		case err != nil:
-			slog.Error("KV GetString failed", "key", key, "error", err)
-		case data == "" || data == kvNullSentinel:
-			// Key not found — fall through to factory().
-		default:
+		if err == nil && data != "" && data != kvNullSentinel {
 			restored, uerr := p.unmarshal([]byte(data))
 			if uerr == nil {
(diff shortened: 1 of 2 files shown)


yuta-yoshinaga:
@claude

PR #1722, opened 2026-05-06
feat(klondike): persist undo history across KV restores

yuta-yoshinaga:
## Summary

Refs #1654 — Klondike pilot. The remaining solitaire games (FreeCell, Spider, Pyramid, TriPeaks, Accordion, Yukon, RussianSolitaire, Scorpion, etc.) will follow in subsequent PRs that copy this pattern.

When a Cloudflare Workers session was restored from KV, Klondike's \`UnmarshalJSON\` deliberately set \`history = nil\`, so a player who reloaded a tab mid-game lost the entire undo stack. ADR-0028 originally rejected this on the assumption that storing snapshots would blow KV size, but in practice a typical Klondike hand serializes to ~150 KB of history (well under the 25 MB KV value limit), so storing the snapshot array end-to-end is the simpler fix vs. the event-sourcing approach the issue speculated about.

This PR:

- Adds \`klondikeSnapshotJSON\` + \`MarshalJSON\` / \`UnmarshalJSON\` on \`klondikeSnapshot\` so the unexported snapshot fields can round-trip JSON. Field names match Klondike's existing short-key convention (\`tb\`, \`st\`, \`wa\`, \`fd\`, \`ps\`, \`mc\`, \`sl\`, \`np\`, \`pr\`).
- Adds \`History []*klondikeSnapshot \\\`json:\"hi,omitempty\"\\\`\` to \`klondikeJSON\` and threads it through \`Klondike.MarshalJSON\` / \`UnmarshalJSON\` (history is now restored, not zeroed).
- Extends the existing \`klondikeMaxSliceLen=1000\` defence to cover \`History\` and the per-snapshot \`Stock\`/\`Waste\` slices, so a malicious payload cannot OOM the worker via an oversized undo array.
- Updates ADR-0028: keeps the original "alternative considered" entry but adds a consequence noting the snapshot-array approach was adopted in #1654 after the size measurement.

## Test plan

- [x] \`go test -tags test ./internal/domain/ -run TestKlondike\` (all Klondike tests pass)
- [x] \`go test -tags test ./internal/domain/ -run TestKlondike_Persists\` (3 new tests: round-trip preserves history length and moveCount; Undo on restored game restores exact pre-action state; oversized history payloads rejected)
- [x] \`goimports -w\` on changed files
- [ ] CI: full \`go test\`, \`golangci-lint run ./...\`

### DIFF
--- a/docs/adr/0028-kv-session-persistence.md
+++ b/docs/adr/0028-kv-session-persistence.md
@@ -43,7 +43,7 @@ wrangler.toml の `[env.staging]` セクションで staging/prod の KV を分�
 
 | 案 | 不採用理由 |
 |----|-----------|
-| コマンドリプレイ（イベントソーシング） | シャッフルの非決定性により同じゲーム状態を再現できない |
+| コマンドリプレイ（イベントソーシング） | シャッフルの非決定性により同じゲーム状態を再現できない（#1654 で再検討する余地はあるが、snapshot 配列方式の方が実装コストが低かった） |
 | ドメインフィールドの export | 全26ゲームの構造体フィールドを公開する必要があり、カプセル化を破壊する |
 | Durable Objects | 無料枠の制約が KV より厳しく、デモ用途には過剰 |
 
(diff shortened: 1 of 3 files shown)


yuta-yoshinaga:
@claude

PR #1750, opened 2026-05-07
refactor(pages): migrate ThreeCard to GamePageShell

yuta-yoshinaga:
Refs #1650. Final standard-pattern page (Three Card Poker, single-player vs. dealer). Tests 35/35 pass.


yuta-yoshinaga:
@claude

PR #1763, opened 2026-05-07
feat(i18n,omaha): migrate OmahaCuiPresenter to i18n (#1699 Phase 2)

yuta-yoshinaga:
Refs #1699 Phase 2.

26 JP literals → omaha.json (35 new keys, ja+en symmetric). Smoke tested both locales; `go test` + lint clean.

## Phase 2 progress

| Game | Status |
|---|---|
| Sevens, Pinochle, Napoleon, Bridge | merged |
| Omaha (26 lit) | **this PR** |
| Pineapple (25 lit) | next |
| Daifugo (25 lit) | upcoming |
| Pitch (24 lit) | upcoming |

PR #1764, opened 2026-05-07
feat(i18n,pineapple): migrate PineappleCuiPresenter to i18n (#1699 Phase 2)

yuta-yoshinaga:
Refs #1699 Phase 2.

25 JP literals → pineapple.json (28 new keys, ja+en symmetric). Same shape as Omaha (#1763) since they share the poker-family layout, plus discard-phase keys for the Pineapple-specific section.


yuta-yoshinaga:
@claude

PR #1778, opened 2026-05-08
feat(i18n): migrate Durak/President/Cassino/SpiteAndMalice/PigsTail (#1699 Phase 3 — final)

yuta-yoshinaga:
## Summary — Phase 3 final batch

Final political/strategic-family batch — five CUI presenters migrated to \`i18n.T/Tf\`. **Phase 3 is now complete: 50 of ~50 games migrated.**

| Game | Highlights |
|---|---|
| Durak | trump line + bottom card, stock line, attacker/defender labels, player hand + finished, table header / pair-defended / pair-open / empty, three phase banners, three game-end banners (draw / human-loses / cpu-loses) |
| President | playerLine + finished, revolution banner + suffix, exchanges header + line, table line + empty, action pass/play, CPU action header, game-end with rank entries, prompts. Also moved the four rank-name literals (大統領 / 副大統領 / 副スカム / スカム / 不明) into locale keys so \`presidentRankName\` resolves through i18n |
| Cassino | playerLine, table line + empty, builds header + line (with multi-group join), human + CPU action lines, game-end with score entries, prompt + help, action variants (take with optional sweepSuffix / build / trail) |
| Spite and Malice | per-foundation labels (empty/filled), player header with human/CPU label, goal line + empty, human-hand label / empty / entry, CPU hand line, side labels (empty/filled), stock + completed line, turn line, win banners + 4 hint variants (discard/goal/hand/side) |
| Pig's Tail | header, top-card line, playerLine, CPU actions header + safe/penalty variants, game-end with loser banner + card-count suffix, current-turn prompt |

## Drive-by: 4 latent loader bugs fixed

\`durak\`, \`president\`, \`cassino\`, and \`spiteandmalice\` all had locale files but were missing from the \`files\` slice in \`internal/i18n/i18n.go\`. Only \`pigtail\` was already registered. Added all four.

This is the **last** loader-divergence fix from the migration sweep. Across Phase 3 we caught:

| PR | Games added to loader |
|---|---|
| #1762 (Phase 2) | bridge |
| #1773 (batch B) | calculation, bakersdozen, golf, russiansolitaire |
| #1774 (batch C) | fortythieves |
| #1776 (batch E) | gofish |
| #1777 (batch F) | canasta, sevenbridge |
| #1778 (this PR) | durak, president, cassino, spiteandmalice |

Eleven latent bugs total. **Strong case** for the auto-discovery refactor Gemini suggested back on #1773 — planning that as the immediate follow-up now that migrations are finished.

[...]

PR #1777, opened 2026-05-08
feat(i18n): migrate Cribbage/GinRummy/Nertz/Canasta/SevenBridge (#1699 Phase 3)

yuta-yoshinaga:
## Summary

Rummy/melding-family batch — five CUI presenters migrated to \`i18n.T/Tf\`:

| Game | Highlights |
|---|---|
| Cribbage | header (round + dealer), starter line, pegging total + cards, playerLine with dealerMark, gameEnd, four phase prompts (discard/pegging/show/roundEnd), three show-detail label keys + multi-field detail template covering all six scoring factors (15s/pairs/runs/flush/nobs) |
| Gin Rummy | header, discardLine, playerLine, gameEnd, four phase prompts (draw/discard with knock/layoff with skip/roundEnd) |
| Nertz | header, foundationsHeader + per-foundation labels, playerLine + nertz/tableau/waste lines (each with empty variants), four hint-zone templates (\`nertz\`/\`waste\`/\`tableau\` ± idx/\`foundation\`), hint line, three phase banners |
| Canasta | new \`canasta.json\` (was unregistered), header + frozenTag, discardLine, playerLine + red3s tag + canasta star, three meld-type variants (mixed/natural/canastaSuffix), meldLine, four phase prompts |
| Seven Bridge | new \`sevenbridge.json\` (was unregistered), header, discardLine, playerLine + meldsHeader, three phase prompts (draw + ds/pon/chi helps / play + meld/layoff/discard helps / roundEnd) |

## Drive-by: 2 latent loader bugs fixed

Both \`canasta\` and \`sevenbridge\` had **no** locale files and were missing from the \`files\` slice in \`internal/i18n/i18n.go\`. Created ja/en files for both and registered the namespaces. Same broader pattern as the gofish fix in #1776 — the auto-discovery follow-up Gemini suggested in #1773 is increasingly justified, but doing the migrations first lets the discovery PR run on a quiet baseline.

## Phase 3 progress

47 of ~50 games done. Final batch G (political/strategic): **President, Cassino, SpiteAndMalice, PigsTail, Durak** — all 5 missing from the loader (same pattern). After batch G lands the auto-discovery refactor PR can replace the hand-maintained slice in one focused change.

## Test plan
- [x] \`go test -tags test ./internal/adapter/presenter/ -run 'TestCribbage|TestGinRummy|TestNertz|TestCanasta|TestSevenBridge'\` — all pass
- [x] \`go test -tags test ./...\` — all packages green
[...]

PR #1776, opened 2026-05-08
feat(i18n): migrate OldMaid/GoFish/EgyptianRatscrew/CrazyEights/Shithead (#1699 Phase 3)

yuta-yoshinaga:
## Summary

Second misc-family batch of Phase 3 — five matching/shedding-style CUI presenters migrated to \`i18n.T/Tf\`:

| Game | Highlights |
|---|---|
| Old Maid | Normal vs JijiNuki output titles, finished/playing player lines, draw action (with optional card reveal + discard suffix), CPU action history, draw history with finished markers, MetaAI line, game-end with Jiji-Nuki removed-card suffix, current-turn prompt with optional draw target |
| Go Fish | new \`gofish.json\` (was unregistered), playerLine + deckLine, ask success/fail, book completion, CPU action variants, error line, per-player score entries on win |
| Egyptian Ratscrew | stock lines, board labels, chance-battle prompt, slappable prompt, three slap-label variants (default/pair/sandwich), 8 slap event lines, win/draw banners |
| Crazy Eights | header, discard + chosen-suit suffix, playerLine, game-end, five phase prompts |
| Shithead | outputTitle, playerLine with turn-arrow + finished suffix, hand/faceup labels, discard + empty, stock, seven-active + skip-next notices, CPU actions header, actionPickup vs actionPlay with optional burn/skip suffix, game-end + rank lines (with Shithead suffix), prompts |

## Drive-by: 1 latent loader bug fixed

\`gofish\` was missing from the \`files\` slice in \`internal/i18n/i18n.go\`, and no \`gofish.json\` existed at all (unlike fortythieves/calculation/etc., which had the file but missed the registration). Created both ja/en \`gofish.json\` and registered the namespace. Same broader pattern of loader-divergence that's bitten this codebase repeatedly — auto-discovery follow-up still warranted.

## Caveat carried over from batch D

\`[エラー] {{msg}}\` banner in EgyptianRatscrew remains hardcoded (same hold-out as War/Slapjack/Speed in #1775). Will be unified under \`cuiErrorBlock\` in a follow-up sweep.

## Phase 3 progress

42 of ~50 games done (14 trick-taking + 17 solitaire + 11 misc). Remaining ~8:
- **Batch F** — Cribbage, GinRummy, Nertz, Canasta\*, SevenBridge\* (rummy/melding family; \* = missing from loader)
- **Batch G** — President\*, Cassino\*, SpiteAndMalice\*, PigsTail\*, Durak\* (political/strategic; all missing from loader)

[...]

PR #1775, opened 2026-05-08
feat(i18n): migrate War/FiftyOne/Trash/Speed/Slapjack/Tonk (#1699 Phase 3)

yuta-yoshinaga:
## Summary

First misc-family batch of Phase 3 — six small fast-paced CUI presenters migrated to \`i18n.T/Tf\`:

| Game | Highlights |
|---|---|
| War | per-player stats line, board labels with empty-card sentinel, 3 phase prompts (reveal/resolved/warBury), per-player win banner |
| Fifty-one | new \`outputTitle\` (\"Fifty-one (フィフティワン)\" — \`helpTitle\` already had \"コマンド\" suffix that's wrong for the game header), human/cpu score lines, table header, stop announcement, error line, end-of-game score table |
| Trash | board labels with \"(turn)\" suffix, faceUp/Down slot templates, stock + discard line, pending card, await-wild prompt; reuses \`cuiSolitaireMoves\` |
| Speed | cpu/human stats, center label + card, hint line, stuck prompt, win banners |
| Slapjack | cpu/human stock lines, board card / empty, three turn prompts (jackOnTop / human / cpu), four slap events (correct/wrong x human/cpu) |
| Tonk | round/stock header, discard line, player line, draw/discard phase prompts with helps, knock help, dealt-Tonk announcement, round-end + nextround help |

## Caveat

The \`[エラー] {{msg}}\` line used by War / Slapjack / Speed remains hardcoded — the bracket prefix is fixed and the body is the user-supplied error string, so converting it correctly belongs in a follow-up that unifies all error-banner styles via the existing \`cuiErrorBlock\` helper. Tonk and Trash already use \`cuiErrorBlock\` per the new convention.

## Phase 3 progress

37 of ~50 games done (14 trick-taking + 17 solitaire + 6 misc). Remaining ~13:
- OldMaid, GoFish, EgyptianRatscrew, CrazyEights, GinRummy, Cribbage, Shithead, Nertz (presenters with locale already registered)
- Canasta, SevenBridge, President, Cassino, SpiteAndMalice, PigsTail, Durak (locales exist but **not** registered in the i18n.go files slice — these will need the same loader fix as the prior batches)

## Test plan
- [x] \`go test -tags test ./internal/adapter/presenter/ -run 'TestWar|TestFiftyOne|TestTrash|TestSpeed|TestSlapjack|TestTonk'\` — all pass
- [x] \`go test -tags test ./...\` — all packages green
[...]

PR #1779, opened 2026-05-08
refactor(i18n): auto-discover locale files via fs.ReadDir

yuta-yoshinaga:
## Summary

Replace the hand-maintained \`files\` slice in \`internal/i18n/i18n.go\` with an \`fs.ReadDir\` walk over \`locales/<lang>/\`. Dropping a new locale JSON on disk is now sufficient to register its keys — no separate code edit needed.

## Why now

The hand-maintained slice was the source of repeated loader-divergence bugs. Across the #1699 Phase 3 migration we caught **eleven** locale files that existed on disk but weren't read by the loader because they had been omitted from the slice:

| PR | Files added back to loader |
|---|---|
| #1762 | \`bridge\` |
| #1773 | \`calculation\`, \`bakersdozen\`, \`golf\`, \`russiansolitaire\` |
| #1774 | \`fortythieves\` |
| #1776 | \`gofish\` |
| #1777 | \`canasta\`, \`sevenbridge\` |
| #1778 | \`durak\`, \`president\`, \`cassino\`, \`spiteandmalice\` |

In every case the symptoms were identical and silent: \`i18n.T(\"<game>.helpTitle\")\` returned the literal key string, breaking both CUI titles and Web-presenter \`MessageCode\` lookups. The Gemini reviewer originally suggested this refactor on #1773; we deferred it until migrations were complete so the rebase pain stayed contained.

## What changed

- \`loadTranslations\` now uses \`fs.ReadDir(fsys, \"locales/\"+lang)\` to enumerate \`*.json\` files at load time.
- The 65+ entry \`files\` slice is gone.
- The companion \`globalNamespaces\` map (\`common\`, \`cui_common\`) is **kept** — it controls whether a file's keys merge flat or with a \`<file>.\` prefix. It only changes when a new global namespace is introduced (twice in the lifetime of the package; never the source of a divergence bug).

## Compatibility

| Aspect | Before | After |
|---|---|---|
| \`loadTranslations\` signature | \`(fs.FS, lang)\` | unchanged |
| Per-game key namespace | \`<file>.<key>\` | unchanged |
| Global key namespace | flat | unchanged |
| Malformed JSON handling | skip silently | unchanged |
| Iteration order on collision | slice order | lexical (ReadDir sorted). \`common\` < \`cui_common\` in both, so observably identical. |
| Subdirectories / non-JSON files | implicitly ignored | explicitly ignored |
| Adding a new locale file | edit slice + drop JSON | drop JSON only |

## Tests added

- **\`TestLoadTranslations_AutoDiscovery\`** uses \`fstest.MapFS\` to verify that a previously-unknown game file is picked up automatically, that global namespaces still merge flat, and that non-JSON files and subdirectories are ignored.
- **\`TestLoadTranslations_AllExistingLocaleFilesResolve\`** walks every embedded \`*.json\` and asserts each contributes at least one key, guarding against future regressions of this exact bug class.

The existing \`TestLoadTranslations_MissingFile\`, \`_InvalidJSON\`, and \`_ValidLang\` tests cover the error and happy paths and continue to pass.

## Conflicts with in-flight PRs

This change conflicts with #1776, #1777, and #1778 — each of those adds entries to the slice this PR deletes. The conflict resolution on rebase is mechanical: the slice no longer exists, so the conflict marker can simply be deleted in full (the new locale files those PRs add are picked up by ReadDir without any code change).

Suggested merge order:
1. Merge #1775, #1776, #1777, #1778 (Phase 3 batches D–G) first.
2. Rebase this PR on the resulting develop and resolve the conflict by deleting the slice additions.
3. Merge.

## Test plan
- [x] \`go test -tags test ./internal/i18n/...\` — all pass (existing + 2 new tests)
- [x] \`go test -tags test ./...\` — all packages green
- [x] \`golangci-lint run ./...\` — 0 issues
- [x] \`goimports -w internal/i18n/i18n.go internal/i18n/i18n_internal_test.go\`

### DIFF
--- a/internal/i18n/i18n.go
+++ b/internal/i18n/i18n.go
@@ -84,12 +84,6 @@ var globalNamespaces = map[string]bool{
 // the entries into a single map. Per-game files are namespaced as
 // "<file>.<key>"; files in globalNamespaces (common, cui_common) are merged
 // flat so their keys are reusable across the codebase.
-//
-// Discovery is automatic via fs.ReadDir over the embedded locales tree —
-// dropping a new locales/<lang>/<game>.json on disk is sufficient to
-// register it. The previous hand-maintained slice repeatedly drifted out
-// of sync with the on-disk files (eleven latent bugs caught in #1699
-// Phase 3 alone), so the slice is gone for good.
 func loadTranslations(fsys fs.FS, lang string) map[string]string {
 	result := map[string]string{}
 	entries, err := fs.ReadDir(fsys, "locales/"+lang)
--- a/internal/i18n/i18n.go
+++ b/internal/i18n/i18n.go
@@ -80,16 +80,32 @@ var globalNamespaces = map[string]bool{
 	"cui_common": true,
 }
 
+// loadTranslations reads every *.json file under locales/<lang>/ and merges
+// the entries into a single map. Per-game files are namespaced as
+// "<file>.<key>"; files in globalNamespaces (common, cui_common) are merged
+// flat so their keys are reusable across the codebase.
+//
+// Discovery is automatic via fs.ReadDir over the embedded locales tree —
+// dropping a new locales/<lang>/<game>.json on disk is sufficient to
+// register it. The previous hand-maintained slice repeatedly drifted out
+// of sync with the on-disk files (eleven latent bugs caught in #1699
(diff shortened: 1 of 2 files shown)


yuta-yoshinaga:
@claude

PR #1780, opened 2026-05-08
refactor(i18n): unify last 4 hardcoded [エラー] banners under cuiErrorBlock

yuta-yoshinaga:
## Summary

War, Slapjack, and Speed each had:

\`\`\`go
fmt.Fprintf(b, \"%s %s\\n\", color.Red(\"[エラー]\"), lastErr.Error())
\`\`\`

while every other presenter in the codebase had already converged on the shared \`cuiErrorBlock(b, lastErr)\` helper that color-reds only the error body. This was the last hold-out from the Phase 3 \"caveat carried over\" note in #1775/#1776 — a follow-up sweep was always the plan once the migrations landed.

FiftyOne was a fifth variant: it had its own \`fiftyone.errorLine\` key (\`\"[エラー] {{msg}}\"\` / \`\"Error: {{msg}}\"\`). Rather than keeping a per-game key for what is essentially a shared concern, this PR drops it and routes through cuiErrorBlock too.

## What changed

| File | Change |
|---|---|
| WarCuiPresenter.go | 3-line error block → \`cuiErrorBlock(b, lastErr)\`; \`fmt\` import removed |
| SlapjackCuiPresenter.go | same |
| SpeedCuiPresenter.go | same |
| FiftyOneCuiPresenter.go | replace \`i18n.Tf(\"fiftyone.errorLine\", \"msg\", lastErr.Error())\` with \`cuiErrorBlock(b, lastErr)\` |
| ja/fiftyone.json | drop \`errorLine\` key |
| en/fiftyone.json | drop \`errorLine\` key |
| 4 \*_test.go files | drop \`assert.Contains(t, result, \"[エラー]\")\` (cuiErrorBlock has no bracket prefix); body-text assertions are kept |

## Net effect

- Zero hardcoded JP error labels in \`internal/adapter/presenter/\`.
- One game-specific i18n key removed from both locales.
- \`fmt\` import removed from 3 presenters (auto-dropped by goimports).

A forbidigo lint rule (Phase 4) can now sit on top of a clean baseline.

## Test plan
- [x] \`go test -tags test ./internal/adapter/presenter/ -run 'TestWar|TestSlapjack|TestSpeed|TestFiftyOne'\` — all pass
- [x] \`go test -tags test ./...\` — all packages green
- [x] \`golangci-lint run ./...\` — 0 issues
- [x] \`grep -rln 'color\.Red(\"\\[エラー\\]\")' internal/adapter/presenter/\` — empty
- [x] \`grep -rln '\"errorLine\"' internal/i18n/locales/\` — empty

### DIFF
--- a/frontend/src/components/BettingControls.test.tsx
+++ b/frontend/src/components/BettingControls.test.tsx
@@ -172,9 +172,13 @@ describe('BettingControls', () => {
   it('applies error styling to input when out of range', () => {
     render(<BettingControls {...makeProps({ betAmount: 80, maxBetAmount: 50 })} />);
     const input = screen.getByLabelText('ベット額:');
-    expect(input.className).toContain('bg-ds-error/40');
+    // Background stays on surface; text stays on text-ds-text-primary (10.1:1 AAA).
+    // Pairing text-ds-error with bg-ds-surface only hits ~2.7:1 — fails AA — so
+    // the error semantic comes from the coloured border, not the text colour.
+    // See `fixup(a11y): keep error/info badge text on text-ds-text-primary for AAA`.
+    expect(input.className).toContain('bg-ds-surface');
     expect(input.className).toContain('border-ds-error');
-    expect(input.className).toContain('text-ds-error');
+    expect(input.className).toContain('text-ds-text-primary');
   });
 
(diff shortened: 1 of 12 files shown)


yuta-yoshinaga:
@claude

PR #1758, opened 2026-05-07
feat(i18n): introduce cui_common locale + i18n the shared CUI helpers (#1699 Phase 1)

yuta-yoshinaga:
Refs #1699 (Phase 1 / Foundation).

The 71 `*CuiPresenter.go` files contain ~1,589 hardcoded Japanese lines and ~563 unique JP literals, so `trumpcards --lang en <game>` advertises a feature that the implementation does not deliver. This PR lays the groundwork for per-game migrations (Phases 2/3) **without changing user-facing behaviour at the default `ja` locale**.

## Changes

- **`internal/i18n/locales/{ja,en}/cui_common.json`** (new) — high-frequency phrases shared across CUI presenters:
  - Player names: `cuiPlayerYou`, `cuiPlayerCpu`, `cuiPlayerUnknown`
  - Betting actions: `cuiBettingAction{Fold,Check,Call,Bet,Raise,AllIn,Unknown}`
  - Trick line: `cuiTrickLine`
  - Shared hint reasons: `cuiHint{FollowSuit,LeadLow,LeadStrong,DiscardHigh,TrumpCut,StrategicBid,StrongHand,WeakHand}`
- **`internal/i18n/i18n.go`** — `loadTranslations` learns a `globalNamespaces` set so `cui_common.json` joins `common.json` in the unprefixed namespace. Every new key uses a `cui*` prefix so collisions with `common.json` are impossible by construction.
- **`cui_card_helper.go`** — `cuiPlayerName` and `cuiBettingActionName` now read from i18n. The hardcoded `bettingActionNames` Japanese map is replaced by `bettingActionKeys` (action → i18n key).
- **`cui_output_helper.go`** — `cuiTrickBlock` and `lookupHintReason` now read from i18n. `sharedHintReasons` becomes `sharedHintReasonKeys` for the same reason.
- **Tests** — new assertions for the English path on every helper plus loader guarantees (`cuiPlayerYou` / `cuiBettingActionFold` / `cuiHintFollowSuit` round-trip in both locales). Existing assertions using Japanese literals continue to pass because the default locale is still `ja`.

## Behaviour at the boundary

Game-specific hint-reason maps (`gameReasons` argument to `lookupHintReason`) are **not** i18n-aware yet — they still pass through verbatim. A new test documents this so the boundary between Phase 1 (helpers) and Phase 2/3 (per-game presenters) is explicit.

## Test plan

[...]

PR #1781, opened 2026-05-08
test(i18n): guard CuiPresenter sources against hardcoded JP string literals (#1699 Phase 4)

yuta-yoshinaga:
## Summary

Closes the i18n migration loop with an automated regression guard.

A new \`TestNoJapaneseStringLiteralsInCuiPresenters\` walks every
\`internal/adapter/presenter/*CuiPresenter.go\` source file via
\`go/ast\` and fails if any string literal contains hiragana, katakana,
half-width katakana, or CJK ideographs. The check is scoped to the
CUI presentation layer only — Go comments, test files, Web presenters,
controllers, and infrastructure are unaffected.

## Why an AST test instead of forbidigo

\`golangci-lint\`'s \`forbidigo\` matches function-call expressions, not
string-literal contents. To enforce \"no JP chars in string literals
under a specific directory,\" the cleanest fit is a tiny repo-local
test that uses \`go/ast\` to inspect \`*ast.BasicLit\` nodes. ~50 lines,
no new lint config, no new dependencies, runs as part of the existing
\`go test ./...\` flow.

## Why the migration matters

Phase 3 (#1699) caught **eleven** loader-divergence bugs — locale files
present on disk but unread by the loader, silently rendering keys as
the literal key string. The auto-discovery refactor (#1779) eliminated
the disk → loader gap; this test closes the source-code → i18n gap.
After this PR a future presenter that inlines a JP string instead of
calling \`i18n.T/Tf\` trips the test at PR time, not in production.

## Drive-by: 3 Skat baseline cleanups

The migration sweep missed three hardcoded JP literals in
\`SkatCuiPresenter.go\`. Cleaned up here so the new test passes from
day one:

| Line | Before | After |
|---|---|---|
| 46 | \`buildCuiOutput(\"Skat (スカート)\", …)\` | \`buildCuiOutput(i18n.T(\"skat.helpTitle\"), …)\` |
| 90 | \`\"b 0/1 ・・・pass (0) or accept …\"\` | \`\"b 0/1 - pass (0) or accept …\"\` |
| 93 | \`\"ps 0/1 ・・・decline (0) or pick up …\"\` | \`\"ps 0/1 - decline (0) or pick up …\"\` |

The first one routes through the existing \`skat.helpTitle\` locale
key (no JSON change needed). The other two used a full-width katakana
middle dot (U+30FB) as a visual separator inside otherwise-English
prompts; replaced with ASCII \`-\` to stay consistent with the rest of
Skat's English CUI output.

## Verification

- \`go test -tags test ./internal/adapter/presenter/ -run TestNoJapanese...\` passes on the cleaned-up baseline.
- Injected \`\"テスト\"\` into a CuiPresenter → test correctly fails with file:line:literal.
- \`go test -tags test ./...\` — all green.
- \`golangci-lint run ./internal/adapter/presenter/...\` — 0 issues.

## Phase 3 / 4 retrospective

| Phase | What it shipped |
|---|---|
| 1 | cui_common helpers (#1758) |
| 2 | 8 trick-taking presenter migrations |
| 3 | 50 game presenters migrated across 14 PRs (#1758-#1778) |
| Auto-discovery | Replaced hand-maintained \`files\` slice with \`fs.ReadDir\` (#1779) |
| Error unification | Last 4 hardcoded \`[エラー]\` banners → \`cuiErrorBlock\` (#1780) |
| **Phase 4 (this PR)** | **Automated regression guard** |

After this lands, the i18n migration is fully closed-loop: every CUI
presenter resolves through i18n, every locale file is auto-discovered,
every error banner uses the shared helper, and every regression is
caught at PR time.

### DIFF
--- a/internal/adapter/presenter/SevensCuiPresenter.go
+++ b/internal/adapter/presenter/SevensCuiPresenter.go
@@ -141,7 +141,10 @@ func (p *SevensCuiPresenter) Output(s interfaces.SevensGame, lastErr error) stri
 	mins := s.GetTableMinVals()
 	maxs := s.GetTableMaxVals()
 	for i, suit := range suits {
-		fmt.Fprintf(&b, "  %s: %d〜%d\n", suitNames[i], mins[suit], maxs[suit])
+		b.WriteString(i18n.Tf("sevens.boardSuitRange",
+			"suit", suitNames[i],
+			"min", strconv.Itoa(mins[suit]),
+			"max", strconv.Itoa(maxs[suit])) + "\n")
 	}
 
 	// Human's previous action
--- a/internal/adapter/presenter/SkatCuiPresenter.go
+++ b/internal/adapter/presenter/SkatCuiPresenter.go
@@ -7,6 +7,7 @@ import (
 	"github.com/yuta-yoshinaga/go_trumpcards/internal/color"
 	"github.com/yuta-yoshinaga/go_trumpcards/internal/domain"
 	"github.com/yuta-yoshinaga/go_trumpcards/internal/domain/interfaces"
+	"github.com/yuta-yoshinaga/go_trumpcards/internal/i18n"
 )
 
 // skatPlayerStr returns the display string for a single Skat player.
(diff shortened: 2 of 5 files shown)


yuta-yoshinaga:
@claude

PR #1782, opened 2026-05-09
refactor(i18n): route VideoPokerCuiPresenter output through i18n (#1699)

yuta-yoshinaga:
Pilot for the final batch of `*CuiPresenter.go` i18n migrations under #1699.

10 / 71 presenters were still bypassing i18n. They squeaked past the lint guard added in #1781 only because their string literals happened to be ASCII-only English (with JP comments). For Japanese users this meant the game body never localized — `--lang ja` showed `phase: BET` / `chips: 1000` / `[HOLD]` / `Four of a Kind! You win!` verbatim.

Migrating them one PR at a time. **VideoPoker is the smallest (80 lines, no shared helpers)** and was chosen as the pilot to lock in the pattern before tackling the larger ones.

## Changes

- **`internal/i18n/locales/{ja,en}/videopoker.json`** — added 12 keys covering every presenter literal: `chipsLine` / `phaseLine` / `phase{Bet,Draw,Result,Unknown}` / `handHeader` / `holdLabel` / `betLine` / `winLine` / `noWin` / `payoutLine`. EN side preserves the previous wording byte-for-byte (`chips: 1000`, `phase: BET`, `[HOLD]`, `bet: 3 coin(s)`, `Four of a Kind! You win!`, `No winning hand.`, `payout: 25`) so the English UX is unchanged.
- **`internal/adapter/presenter/VideoPokerCuiPresenter.go`** — routed `Output()` and `phaseStr()` through `i18n.T` / `i18n.Tf` with `strconv.Itoa` for int substitutions.
- **`internal/adapter/presenter/VideoPokerCuiPresenter_test.go`** — updated assertions to expect the new JA output. The Go i18n module defaults to ja in tests (`i18n.go:124` pre-loads ja), matching the convention already used by `PageOneCuiPresenter_test.go` and other migrated presenters.

## Pattern (for the remaining 9)

1. Add ja + en keys to `internal/i18n/locales/{ja,en}/<game>.json` — keep en byte-identical to the current literal output.
2. Replace literals with `i18n.T("<game>.key")` / `i18n.Tf("<game>.key", "param", value)`.
3. Use `strconv.Itoa(n)` (not `fmt.Sprintf("%d", n)`) for int → string when feeding `Tf`.
4. Update tests to assert on the new ja output.

## Verification

- `go test -tags test ./internal/adapter/presenter/...` — all pass, incl. `TestNoJapaneseStringLiteralsInCuiPresenters`
- `golangci-lint run ./internal/adapter/presenter/...` — 0 issues
- `goimports -w` applied on touched Go files

## Remaining for #1699

9 ASCII-only presenters: Baccarat, BlackJack, CaribbeanStud, CasinoWar, LetItRide, PaiGow, RedDog, TexasHoldemBonus, ThreeCard. Each will follow as its own small PR with the same pattern.

Refs #1699

### DIFF
--- a/internal/adapter/presenter/VideoPokerCuiPresenter.go
+++ b/internal/adapter/presenter/VideoPokerCuiPresenter.go
@@ -2,11 +2,13 @@ package presenter
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 
 	"github.com/yuta-yoshinaga/go_trumpcards/internal/color"
 	"github.com/yuta-yoshinaga/go_trumpcards/internal/domain"
 	"github.com/yuta-yoshinaga/go_trumpcards/internal/domain/interfaces"
+	"github.com/yuta-yoshinaga/go_trumpcards/internal/i18n"
 )
 
 // VideoPokerCuiPresenter ビデオポーカーCUIプレゼンタークラス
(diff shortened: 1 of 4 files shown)


yuta-yoshinaga:
@claude

PR #1792, opened 2026-05-09
refactor(cui): route IndianPokerCuiPresenter through buildCuiOutput (#1701)

yuta-yoshinaga:
Pilot for #1701 (CuiPresenter shell-helper consolidation).

## Scope correction vs. issue text

The issue describes 71 presenters with hand-rolled shells. Today on develop the actual state is:

| | |
|---|---|
| **52** of 71 | already use the existing `buildCuiOutput` helper (`internal/adapter/presenter/cui_output_helper.go`) ✓ |
| **9** of 71 | still hand-roll \`==========\n<title>\n==========\n...\n==========\n\` |
| **10** of 71 | use a different `----------` separator shape (the casino games — out of scope for this helper) |

So the actionable scope is **9 holdouts**, all poker-family (Badugi, Holdem, IndianPoker, Omaha, Pineapple, Poker, Sevens, ShortDeck, SevenCardStud). One PR per game, smallest first. **IndianPoker (117 lines) is the pilot.**

## Changes

- **`internal/adapter/presenter/IndianPokerCuiPresenter.go`** — wrapped `Output()` in `buildCuiOutput(i18n.T("indianpoker.outputTitle"), ...)`, removed the 3 hand-written `==========` writes, switched the error block to the existing `cuiErrorBlock(b, lastErr)` helper. `fmt` import becomes unused and is dropped.

## Visible output change

A trailing `==========\n` footer is now printed (the helper always closes with one). Previously this presenter ended without a footer — anomalous compared to the 52 migrated peers (Hearts, Spades, Spider, Klondike, etc.). Bringing it into alignment is part of the consolidation point.

Before:
\`\`\`
==========
Indian Poker
==========
<body>
[CPU行動]
  Player 1: チェック
  ...
\`\`\`
After:
\`\`\`
==========
Indian Poker
==========
<body>
[CPU行動]
  Player 1: チェック
  ...
==========
\`\`\`

## Verification

- `go test -tags test ./internal/adapter/presenter/...` — pass, incl. `TestNoJapaneseStringLiteralsInCuiPresenters`
- `golangci-lint run ./internal/adapter/presenter/...` — 0 issues
- `goimports -w` applied
- Manual smoke test via `go run ./cmd/trumpcards` confirms the shell renders correctly.

## Remaining for #1701

8 hand-rolled-shell presenters: Badugi, Holdem, Omaha, Pineapple, Poker, Sevens, ShortDeck, SevenCardStud.

[...]

PR #1793, opened 2026-05-09
refactor(cui): route BadugiCuiPresenter through buildCuiOutput (#1701)

yuta-yoshinaga:
Second of the 9 hand-rolled-shell migrations under #1701 (after #1792 IndianPoker pilot). Same mechanical migration: wrap `Output()` in `buildCuiOutput`, use `cuiErrorBlock` for error rendering, drop the now-unused `fmt` import.

## Visible change

Trailing `==========\n` footer is now present (matching the 52 already-migrated peers). Previously this presenter ended without a footer.

## Verification

- `go test -tags test ./internal/adapter/presenter/...` — pass, incl. `TestNoJapaneseStringLiteralsInCuiPresenters`
- `golangci-lint run ./internal/adapter/presenter/...` — 0 issues

## Remaining for #1701

7 hand-rolled-shell presenters: Holdem, Omaha, Pineapple, Poker, Sevens, ShortDeck, SevenCardStud.

Refs #1701

### DIFF
--- a/internal/adapter/presenter/BadugiCuiPresenter.go
+++ b/internal/adapter/presenter/BadugiCuiPresenter.go
@@ -1,7 +1,6 @@
 package presenter
 
 import (
-	"fmt"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


yuta-yoshinaga:
@claude

PR #1794, opened 2026-05-09
refactor(cui): route PokerCuiPresenter through buildCuiOutput (#1701)

yuta-yoshinaga:
Third of the 9 hand-rolled-shell migrations under #1701 (after #1792 IndianPoker, #1793 Badugi).

## Changes

- `Output()` wrapped in `buildCuiOutput`. Title selection (lowball vs standard) hoisted into a local `titleKey` before the wrap so the helper still receives a single string.
- Error block uses `cuiErrorBlock(b, lastErr)`.
- **`OutputWithOdds()`** adjusted: it previously appended a separator `==========\n` then the odds header. Now that the base `Output()` ends with the `buildCuiOutput` footer (also `==========\n`), the redundant divider is dropped — odds section appends as a continuation.

## Visible change

- Trailing `==========\n` on `Output()` (matches the 52 migrated peers).
- `OutputWithOdds()` net-net: same divider count as before, just rearranged (base footer now serves as the divider, instead of an explicit one in the odds builder).

## Verification

- `go test -tags test ./internal/adapter/presenter/...` — pass, incl. `TestNoJapaneseStringLiteralsInCuiPresenters`
- `golangci-lint` — 0 issues

## Remaining for #1701

6: Holdem, Omaha, Pineapple, Sevens, ShortDeck, SevenCardStud.

Refs #1701

### DIFF
--- a/internal/adapter/presenter/PokerCuiPresenter.go
+++ b/internal/adapter/presenter/PokerCuiPresenter.go
@@ -16,130 +16,123 @@ type PokerCuiPresenter struct{}
 
 // Output renders the current game state for the active locale (#1699).
 func (pcp *PokerCuiPresenter) Output(p interfaces.PokerGame, lastErr error) string {
-	var b strings.Builder
-	players := p.GetPlayers()
-
-	b.WriteString("==========\n")
+	titleKey := "poker.outputTitle"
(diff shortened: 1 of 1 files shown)


yuta-yoshinaga:
@claude

PR #2108, opened 2026-05-30
feat(solitaire): require confirmation before give-up (infra + Klondike)

yuta-yoshinaga:
## Summary

Give-up abandons an in-progress solitaire game and is **irreversible**, yet — unlike reset — it fires immediately from both the give-up button **and** the single `g` key. A mis-tap or stray keypress throws away minutes of progress with no guard (issue #2099). Reset already routes through a confirm dialog; this brings give-up to parity.

This PR lands the **shared infrastructure + the Klondike reference implementation**. The remaining solitaire pages are a mechanical follow-up (see below).

## What's in this PR

- **i18n** (`common.json` ja+en): `button.confirmGiveUp` / `confirmGiveUpMessage` — give-up-specific copy, **not** reused reset copy (投了確認 / "Confirm Give Up").
- **`useGamePageSetup`**: a second, independent `useConfirmDialog` instance exposed as `giveUpConfirmOpen` / `requestGiveUpConfirm` / `confirmGiveUp` / `cancelGiveUp`, so give-up and reset dialog states never collide.
- **`GameGiveUpDialog`**: mirrors `GameResetDialog` with the give-up keys.
- **`GamePageShell`**: optional `giveUpConfirmOpen`/`confirmGiveUp`/`cancelGiveUp` props; renders the dialog only when provided, so non-solitaire pages are untouched.
- **KlondikePage** (reference wiring): `confirmGiveUpAction = () => requestGiveUpConfirm(handleGiveUp)` now backs both the button `onClick` and the `g` keybinding; the shell receives the give-up props. Confirm dispatches `giveup`; cancel does nothing.

## Follow-up (intentionally not in this PR)

The other ~24 solitaire pages need the same 5-line wiring. I attempted a scripted bulk rollout but it produced non-trivial per-page issues (use-before-declaration on pages that define `handleGiveUp` locally, hook dep-array updates, and three pages — Spiderette/SeahavenTowers/BeleagueredCastle — that render two `GamePageShell` instances and so double-mount the dialog). Rather than land churn I couldn't fully verify in this session, I'm shipping the verified vertical slice and will wire the rest page-by-page in a follow-up. Tracked on #2099.

## Test plan
- `GameGiveUpDialog` test: renders give-up copy; confirm/cancel callbacks fire.
- `useGamePageSetup` test: give-up dialog state is independent of reset; cancel runs no action.
- `KlondikePage` test: give-up does **not** dispatch before confirm; confirm dispatches `giveup`; cancel dispatches nothing.
- `bun run test -- --run KlondikePage GameGiveUpDialog useGamePageSetup GamePageShell GameResetDialog` ✅ 5 files / 119 tests.
- `bunx biome check` on all 9 changed files ✅.
- tsc type-check runs in CI.

Refs #2099 (infra + Klondike; remaining pages follow up).

### DIFF
--- a/frontend/src/components/GameGiveUpDialog.test.tsx
+++ b/frontend/src/components/GameGiveUpDialog.test.tsx
@@ -1,0 +1,31 @@
+import { fireEvent, render, screen } from '@testing-library/react';
+import { describe, expect, it, vi } from 'vitest';
+import { GameGiveUpDialog } from './GameGiveUpDialog';
+
+describe('GameGiveUpDialog', () => {
+  it('renders nothing when closed', () => {
+    render(<GameGiveUpDialog giveUpConfirmOpen={false} confirmGiveUp={vi.fn()} cancelGiveUp={vi.fn()} />);
+    expect(screen.queryByRole('alertdialog')).not.toBeInTheDocument();
+  });
+
+  it('renders give-up-specific copy (not reset copy) when open', () => {
+    render(<GameGiveUpDialog giveUpConfirmOpen={true} confirmGiveUp={vi.fn()} cancelGiveUp={vi.fn()} />);
+    // ja translations are loaded in test/setup.ts.
+    expect(screen.getByText('投了確認')).toBeInTheDocument();
+    expect(screen.getByText('投了すると現在のゲームは終了します。よろしいですか？')).toBeInTheDocument();
+  });
+
+  it('fires confirmGiveUp on confirm click', () => {
+    const confirmGiveUp = vi.fn();
+    render(<GameGiveUpDialog giveUpConfirmOpen={true} confirmGiveUp={confirmGiveUp} cancelGiveUp={vi.fn()} />);
+    fireEvent.click(screen.getByRole('button', { name: '確認' }));
+    expect(confirmGiveUp).toHaveBeenCalledTimes(1);
+  });
+
+  it('fires cancelGiveUp on cancel click', () => {
+    const cancelGiveUp = vi.fn();
+    render(<GameGiveUpDialog giveUpConfirmOpen={true} confirmGiveUp={vi.fn()} cancelGiveUp={cancelGiveUp} />);
+    fireEvent.click(screen.getByRole('button', { name: 'キャンセル' }));
+    expect(cancelGiveUp).toHaveBeenCalledTimes(1);
+  });
+});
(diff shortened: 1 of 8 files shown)


yuta-yoshinaga:
@claude

yuta-yoshinaga:
Thanks both — addressed the two WANT items in `b057f5c8`:

1. **Keyboard guard during auto-complete** (`KlondikePage.tsx`) — `useActionKeyboardNav`'s `enabled` now includes `!isAutoCompleting`, matching the give-up button's `disabled={loading || isAutoCompleting}`. A `g` keypress mid-auto-complete no longer opens the confirm dialog.

2. **Type-unsafe give-up prop group** (`GamePageShell.tsx`) — refactored `GamePageShellProps` into a discriminated union (`GamePageShellBaseProps & GiveUpConfirmProps`) so `giveUpConfirmOpen` / `confirmGiveUp` / `cancelGiveUp` are all-or-nothing. Incomplete wiring is now a compile error instead of a silent no-op dialog — exactly what we want before the ~24-page follow-up rollout.

Notes:
- I did **not** add a dedicated "`g` does nothing while `isAutoCompleting`" test: `isAutoCompleting` is internal `useAutoCompleteState` machine state driven by the auto-complete animation, so forcing it true in a page test is timing-flaky. The code change mirrors the button's already-tested disabled condition. Happy to add one if you'd prefer a hook-level test.
- **Dead-state (INFO)**: leaving `useGamePageSetup`'s second `useConfirmDialog` as-is for now — it's harmless and keeps the call site uniform across pages; can extract a `useGiveUpConfirmDialog()` later if it matters.

Affected tests green (`GamePageShell`, `KlondikePage`, `GameGiveUpDialog`, `useGamePageSetup` — 126 tests); biome clean; tsc runs in CI.

yuta-yoshinaga:
Correction to my previous comment: the discriminated-union refactor didn't actually land in `a66ce4b4` — the edit adding `GiveUpConfirmProps` + `export type GamePageShellProps = GamePageShellBaseProps & GiveUpConfirmProps` silently failed, so that commit left the component signature referencing an **undefined** `GamePageShellProps` (vitest still passed because esbuild erases types without checking, but CI `tsc` would have failed).

Fixed for real in **`9cb1f3e6`** — the union is now defined and exported, the give-up trio is genuinely all-or-nothing, and `bunx biome check` + the affected suites (`GamePageShell`/`KlondikePage`/`GameGiveUpDialog`/`useGamePageSetup`, 115 tests) pass. Sorry for the noise.

Issue #2126, opened 2026-06-03
Split domain/usecase/adapter layers per worker category to fit the 1 MB Cloudflare Worker limit


yuta-yoshinaga:
## Problem

The **classic** Cloudflare Worker is over the 1 MB gzip free-tier limit and has been failing on `develop` since the Schnapsen merge (#2124) — independent of any single feature.

| develop commit | game | classic gzip | vs 1,048,576 B limit |
|---|---|---|---|
| `39340d83` | #121 Five Hundred | 1,046,926 B | 1,650 B under ✅ |
| `9285d6c0` | #122 Schnapsen (merge) | 1,050,209 B | 1,633 B **over** ❌ |

Schnapsen — a **solo** game — added **+3,283 B to the classic worker**. Every new game (any bucket) now grows all three workers, so the workers scale unbounded with the catalog and classic (largest game-logic payload) breaks first.

## Root cause

The casino/classic/solo split only isolates the **infrastructure** layer (per-category blank imports under `js && wasm` build tags). But `internal/domain`, `internal/domain/interfaces`, `internal/usecase`, `internal/adapter/controller`, and `internal/adapter/presenter` are each **monolithic single packages**. `games/classic` imports all of them, and Go compiles a package in full, so the classic worker compiles **every** game's files across all five layers.

TinyGo dead-code-eliminates unreferenced **functions** (game logic), but it cannot eliminate, and these ship in *every* worker:
1. `json.Marshaler` / `json.Unmarshaler` implementations (retained for interface dispatch),
2. package-level `var x = map[...]{...}` initializers (run at each worker's package init),
3. type descriptors for every domain type.

## Proposed fix — split per category

Make each worker compile only its category's game code + shared "core" code.

**Technique A — build tags (recommended, least churn).** Tag each game's files across all 5 layers with `//go:build !js || !wasm || <category>` and add `-tags $(1)` to the Makefile's TinyGo build (`$(1)` is already the worker/category name). Shared helpers (Card, TrumpCards, GamePlayer, betting, hand_eval, …) stay untagged "core". No reference changes — types stay `domain.X`. ~1,000–1,200 one-line additions, scriptable from the registry category map.

**Technique B — sub-packages.** Physically move files into `domain/{core,casino,classic,solo}` (and per-layer equivalents) and rewrite every `domain.X` → `<cat>.X` reference. Cleaner conceptually, but thousands of reference edits + circular-import risk.

## Risks / notes

[...]

PR #2127, opened 2026-06-03
chore(workers): pilot per-category build tags (#2126)

yuta-yoshinaga:
Pilot for #2126 — prove the per-category build-tag split shrinks the classic worker.

- Makefile: TinyGo build now passes `-tags <category>` (the worker name).
- Three **zero-coupling** solo games (Pyramid, TriPeaks, Wasp — nothing references them) carry `//go:build !js || !wasm || solo`, so they compile only in the solo worker + non-worker builds and drop out of classic & casino.

**What to read off CI:** classic gzip should fall ~3 KB/game (~9 KB) below its current ~1,050,200 B; solo/casino/server/test builds unchanged. If validated, extend to the whole solo (then casino) category for the real fix.

Refs #2126

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,14 @@ define build_worker
 	# runCpuTurns → Output on staging, so we bump to 128KB. The flag only
 	# reserves linear memory at runtime (binary stays well under the
 	# Cloudflare Workers free-tier 1MB gzipped limit).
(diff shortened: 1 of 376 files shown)


yuta-yoshinaga:
@claude

PR #2139, opened 2026-06-05
chore: add commit-time game-count assertion guard hook

yuta-yoshinaga:
## Summary

Adds a guard against the single most-recurring new-game first-push failure: a stale hardcoded game-count assertion that slips past local checks and only fails in the expensive, OOM-prone CI round-trip.

Two commits:

1. **`settings.json` PreToolUse hook** — on `git commit`, when staged files touch count-affecting paths (`registry.go`, `registry_test.go`, the `casino/classic/solo` worker sub-packages, or the tutorial-progress assertion tests), it runs the count-audit and **blocks the commit** if the assertions disagree with the `RegisterKVGame` source of truth. Has an existence guard so it degrades to a no-op where the script is absent.
2. **`scripts/count-audit.sh` + SKILL.md wiring** — a read-only audit comparing the per-category `RegisterKVGame` count against every hardcoded assertion: the Go consts in `registry_test.go` **and** the three frontend assertions that are **tsc-only** (`bun run check` / biome does not catch a stale count). Exits non-zero on mismatch. Wired into the `new-game` skill as both the count-bump driver and a pre-commit gate.

## Why these files live under `.claude/`

This is agent tooling. `new-game/SKILL.md` was already tracked; the script and the hook are committed alongside it so the protection travels with the repo rather than relying on a local-only file.

## Test plan

- [x] Hook no-op paths return `{}` (non-commit; commit with nothing relevant staged; commit with count file staged + counts consistent)
- [x] Hook blocking path: staged `registry_test.go` with `expectedSolo 44→45` mismatch → `continue: false` with the exact stale-assertion line; valid JSON output (emoji/newlines escaped via `jq -n`)
- [x] `count-audit.sh` reports all ✅ against current repo (124 games = 47 casino + 33 classic + 44 solo)
- [x] Temporary test mismatch fully reverted; no source files changed

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -24,6 +24,12 @@
             "command": "jq -r '.tool_input.command' | { read -r cmd; case \"$cmd\" in git\\ commit*) echo '{\"hookSpecificInstructions\": \"This is a git commit. Check for documentation drift: run git diff --cached --name-only to see staged files, then verify: (1) game domain/interactor/controller/presenter changes have README.md, CLAUDE.md Commands, docs/games.md, docs/manual/ updates; (2) Web API route changes have docs/architecture.md, api/openapi.yaml updates; (3) frontend page/component changes have frontend/CLAUDE.md i18n list updates; (4) new ADR files have docs/adr/README.md index updates; (5) controller schema changes have api/openapi.yaml updates. Block if required docs are missing from staged files.\"}' ;; *) echo '{}' ;; esac; }",
             "timeout": 10,
             "statusMessage": "Checking for documentation drift..."
+          },
+          {
+            "type": "command",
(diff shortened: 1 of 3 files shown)


yuta-yoshinaga:
@claude

yuta-yoshinaga:
Thanks @gemini-code-assist and @claude — both flagged the same `grepval` fragility. Addressed in 18d4113:

- **`grepval`**: dropped the line-number flag and added `sed 's|//.*||'` to strip inline comments before extracting digits, so `expectedSolo = 44 // up from 39` now reads `44` (previously extracted `39`/`10`).
- **`panel_links` / `panel_incomplete`**: now call `grepval` instead of duplicating the fragile pattern, so they can't diverge.
- **`panel_text`**: added the comment-strip and documented the "total is the largest `getByText(/N/)` literal" assumption in a comment.

Verified: clean run still reports all ✅ (124 games), and a regression test injecting `expectedSolo = 44 // up from 39 in 2024-08-10` correctly reads `44` (the old logic returned `10`).

PR #2927, opened 2026-06-28
feat(catchten): add Catch the Ten (Scotch Whist) trick-taking game

yuta-yoshinaga:
## Summary

Implements **Catch the Ten (Scotch Whist)** (#2888), a Scottish 36-card (ranks 6–A) 4-player partnership trick-taker (game #177, classic bucket).

The trump suit uses the special Scotch Whist order **J(11) > A > K > Q > 10 > 9…6**; non-trump suits are A-high. Whoever wins a trick captures its trump **honours** — J=11, 10=10, A=4, K=3, Q=2. Must-follow tricks. Deals repeat until a team reaches the target (default **41**); if both reach it in the same deal the higher total wins, an exact tie is a draw. Mechanics modelled on the existing `Whist` game.

## Implementation

- **domain** — `CatchTen` two-team state machine, special trump ranking, honour-capture scoring, deterministic game-end/draw tiebreak. `UnmarshalJSON` validates config/phase/trump/seat indices and bounds slice lengths.
- **usecase** — interactor with snapshot/restore for KV persistence.
- **adapter** — CUI + Web controllers and presenters, i18n-routed (ja/en, incl. draw).
- **infrastructure** — registered at all 4 points; **classic** worker (casino is at the 1 MB gzip limit, and as a trick-taker this shares classic's existing trick code). Classic count `54 → 55`, total `176 → 177`.
- **frontend** — page, API client, hint, tutorial, i18n; CLI parser + formatter in **dedicated files** with tests; added to the `games` array and `gameTheme`.
- **docs** — CUI/web manuals, architecture, games, README, openapi.

## Test plan

- [x] Go build clean; domain (39) / usecase / presenter / controller tests pass
- [x] Registry count + doc-consistency + worker-registration tests pass (65/55/57 = 177)
- [x] `goimports` clean; `golangci-lint` 0 issues across 9 packages
- [x] i18n key parity (ja/en) for catchten + common + discover
- [x] Frontend: 34 catchten tests + gameTheme/tutorial/docCount count-guards pass
- [x] `tsc` type-check clean; Biome clean
- [ ] E2E + tinygo classic worker build (run in CI)

Closes #2888

PR #2930, opened 2026-06-29
fix(catchten): stop double-resolving CPU-completed tricks

yuta-yoshinaga:
## Problem

PR #2927 (Catch the Ten) moved trick resolution into `CatchTen.playCard` — calling `ResolveTrick()` as soon as the trick fills — based on a Gemini review comment claiming that a CPU playing the last card of a trick would otherwise never be resolved.

**That premise was incorrect.** The shared `runCpuTurnsLoop` (used by every trick game, including ohhell/whist) *already* calls `ResolveTrick()` after each `CpuPlay`. So a CPU-completed trick was resolved **twice** — once inside `playCard`, then again in the loop (the trick stays full until `NextTrick`, so the guard didn't stop the second call). This **double-counted the winner's tricks and honour points**, and double-ran the final deal's `ScoreRound`.

The existing tests missed it because the domain play-through helper drives tricks directly, not through the interactor's CPU loop where the second resolve happens. Web/E2E don't assert exact scores, so it slipped through.

## Fix

Revert to the established pattern (ohhell / whist):
- `playCard` only marks the trick complete (`phase = TrickEnd`).
- The **caller** resolves exactly once — the interactor's human path in `Play()`, and `runCpuTurnsLoop` for CPU plays.
- The round is scored in `NextRound()` (as before #2927).

## Regression test

`TestCatchTen_PlayCardDoesNotAutoResolve` fills a trick with a **CPU** play and asserts no trick is awarded until a single `ResolveTrick` runs — directly guarding against re-introducing the auto-resolve.

## Verification
- [x] Domain (incl. new regression test), usecase, presenter, controller tests pass
- [x] `golangci-lint` 0 issues (domain, usecase); `goimports` clean

### DIFF
--- a/internal/domain/CatchTen.go
+++ b/internal/domain/CatchTen.go
@@ -253,10 +253,7 @@ func (g *CatchTen) ResolveTrick() {
 	g.leadPlayerIdx = winnerIdx
 
 	if g.trickNumber >= CatchTenHandSize {
-		// Last trick of the deal — tally the round (and the game-end check)
-		// now so the round-end score screen reflects the final scores.
 		g.phase = CatchTenPhaseRoundEnd
-		g.ScoreRound()
 	} else {
 		g.phase = CatchTenPhaseTrickEnd
 	}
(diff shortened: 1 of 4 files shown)


yuta-yoshinaga:
Thanks — but I'm keeping `ScoreRound()` in `NextRound()` (not in `ResolveTrick()`) deliberately, because that is the **established convention** for every trick game in this codebase, not a regression:

- `ohhell` (the template `whist`/`catchten` are cloned from): `ResolveTrick()` only sets `phase = RoundEnd`; scoring happens in `OhHellInteractor.NextRound()` (`internal/usecase/OhHellInteractor.go:102`). Same for `whist`.
- This is exactly what **Catch the Ten did originally**, before PR #2927 — my #2927 change had moved `ScoreRound` into `ResolveTrick`; this PR reverts that along with the buggy `playCard` auto-resolve.

So the round-end screen showing cumulative scores only after "next deal" is consistent with ohhell/whist/and the original catchten — keeping `ScoreRound` in `ResolveTrick` would make catchten the lone outlier. The actual correctness bug (double trick/honour counts on CPU-completed tricks) is fixed by removing the `playCard` auto-resolve, since `runCpuTurnsLoop` already resolves CPU plays. `TestCatchTen_PlayCardDoesNotAutoResolve` guards it.

PR #4338, opened 2026-07-25
feat(frontend): centralize game sound taps with policy layer

yuta-yoshinaga:
## Summary

PR 1 of 2 for the sound-system centralization (eng-reviewed design; PR 2 sweeps the 117 hand-wired pages).

- **SoundProvider policy layer**: per-sound min-interval throttles, exec action-claim token (consumed on next resolution or 3 s expiry), cardPlace burst arpeggio (+3.5 %/step, cap 1.25, 1.5 s idle reset), per-sound `SOUND_ENABLED` kill-switch map, suspended-AudioContext gate (Howler `autoUnlock` replays pre-gesture sounds stale rather than dropping them).
- **useGameApi central tap** (covers all 219 game pages): `reset` command → shuffle (deal AND redeal), other successes → cardPlace unless claimed. Sound context is read via a ref so `exec` keeps its stable identity — 67 files key effects on `[exec]`, so a naive dep would make the mute toggle re-deal games. Sound failures can never break exec or mask API errors.
- **GamePageShell taps**: winFanfare rides the existing celebration trigger (also fixes double fanfares from inline `onCelebrate` identity churn re-arming WinCelebration's timer), lossThud on an opt-in `lossShow` prop (latched) — deliberately NOT derived from `!winShow`, since 22 casino pages pass `winShow={result > 0}` where a push is neither a win nor a loss, turnTick on the `isHumanTurn` rising edge (solitaire pages omitting the prop never tick).
- **BettingControls**: chipClick + exec claim on money actions only (call/raise/bet/all-in).
- **ErrorAlert**: errorBuzz once per new displayed error (~200 pages, zero page wiring).
- **Safari/iOS fix**: `SOUND_FILES` listed `.mp3` fallbacks that didn't exist, so Safari was silent; real mp3s now generated via committed `scripts/generate-sound-fallbacks.sh`.
- **Test infra**: global howler mock in `src/test/setup.ts`; BaccaratPage reveal tests wrapped in `act()` (the stable celebration handler changed scheduler timing under fake timers).
- **Rider**: restores the "currently N" hint-factory count in `frontend/CLAUDE.md` that #4337 dropped — `useGameHint.docCount.test.ts` has been red on develop since.

Temporary: winFanfare/lossThud carry a 3 s dedupe guard while the 100+ page-level sound handlers still exist; PR 2 deletes those handlers, the guard, dead `useHaptics`, and untracks the root `public/sounds/` build output.

## Test plan

- [x] 19 new/updated test cases across SoundProvider (throttle/claim/arpeggio/enable-map/suspended gate), useGameApi (reset→shuffle, cardPlace, claim consumption both branches, providerless render, **exec-identity-across-mute regression**), GamePageShell (fanfare/thud/tick/latch), BettingControls (chip+claim, check/fold excluded), ErrorAlert (buzz once per new error)
- [x] `bun run build` (tsc), `bun run check`, all 8 touched test files green (209 tests)
- [x] Full local suite: 15771+ passing; NavBar.test.tsx worker OOM is a pre-existing local-env issue (reproduced identically on clean origin/develop)
- [ ] CI green
- [x] Auto-review finding (lossThud on pushes/ties) fixed in 91bc8192f — see comment
- [ ] Prod-build ear test (five games) before PR 2 tunes intervals/arpeggio


yuta-yoshinaga:
## Review response — `lossThud` push/tie issue fixed

Good catch, and confirmed: 22 pages pass a `winShow={result > 0}`-shaped condition, so a push (`result === 0`, bet returned) yielded `winShow=false` and would have thudded at break-even. Dragon Tiger and Casino War hit it on every tie.

Rather than leave it for the ear test, fixed in 91bc8192f by removing the derivation entirely:

- New opt-in `lossShow?: boolean` prop on `GamePageShell`; `lossThud` fires only on `lossShow === true`, never on `!winShow`.
- Pages that cannot distinguish loss from push stay silent on game end — the honest v1 behavior, and nothing regresses since `lossThud` had zero pre-existing per-page usage.
- The page sweep (PR 2) wires `lossShow` per game where the result model supports it, alongside the follow-up audit of the ~58 pages that pass no `winShow` at all.
- Regression test added: `winShow: false` with no `lossShow` asserts silence, with the push rationale in the test body.

Re: the review note about `bun run test`/`bun run build` not being runnable in the sandbox — both were run locally: build (incl. tsc) and Biome clean, 29/29 GamePageShell tests, 209 tests across all touched files.

yuta-yoshinaga:
## Second fix: central card sound no longer fires on rejected actions

While preparing the page sweep I found a defect this tap introduced, in the same class as the `lossThud` one: the central `cardPlace` fired on **every** 200 response, including the ones that report a rule rejection (illegal move, out-of-turn play). A misclick in solitaire sounded exactly like a card landing.

Baker's Dozen had already solved this locally — its own sound keys on `moveCount`, so illegal moves (which leave the count unchanged) stay silent, with a comment saying so. The planned sweep would have deleted that precision and left only the beep.

Fixed in 6b2543b8f: skip the sound when the response reports a rejection.

**Why the discriminator is safe** (surveyed all 210 `*WebPresenter.go`):
- Rejections surface as `message = lastErr.Error()` with an **empty** `messageCode` — 199 of 210 presenters (98 direct assignments + 104 `buildMessage` returns of `(msg, "", nil)`).
- **Zero** presenters emit a success-path `message` without an adjacent `messageCode`, so a real state change can never be misread as a rejection.
- The ~11 presenters that do attach a code to a rejection fall through and still sound — the failure mode is today's behavior, never a swallowed legitimate sound.

Tests: rejected response (message, no code) → silent; `message` + `messageCode` (e.g. `bakersdozen.gameClear`) → still plays. 25/25 in `useGameApi.test.ts`, build + Biome clean.

PR #4339, opened 2026-07-25
refactor(frontend): sweep per-page sound wiring into the central taps

yuta-yoshinaga:
## Summary

PR 2 of 2 for the sound-system centralization (PR 1: #4338). PR 1 added the central taps; this deletes the per-page wiring they replaced, so the net diff is negative and every game's audio now comes from one place.

- **Swept 155 of 173 page-level `playSound` sites** across ~120 files, driven by a regenerated per-site manifest rather than judgment calls: 110 `onCelebrate={() => playSound('winFanfare')}` props, 27 move-handler `cardPlace`, 13 reset-handler `shuffle`, 4 error-edge `errorBuzz`. 104 files lose their `useSound` import entirely.
- **Deleted two local workarounds the shell now owns**: CatchTen and TwoTenJack each carried a memoized `handleCelebrate` (with a comment explaining the double-fanfare bug) — PR 1's stable handler makes them redundant.
- **Fixed 3 pages that double-fired today**: Bourré, Dou Dizhu and Tichu ran a `humanWon` effect *and* passed `onCelebrate`, so a win played two fanfares.
- **Kept 12 sites the central taps cannot infer** — documented in the sweep table:
  - *wrong-sound class* (the tap maps only `reset`→shuffle): Monte Carlo's `deal` (regrid), Cruel's `shift` (recompact), Prší's stock draw.
  - *event/animation flavor*: TriPeaks peak-cleared, Spider suit-completed, War phase beats, Indian Poker reveal beat, Basra capture and Belote declaration (both reuse `winFanfare` mid-game), Mao's `rulePenalty` buzz, Durak's take-cards buzz, Spoons' grab-window tick, plus all `cardDeal`/`cardFlip`/`cardSelect` and the bespoke `chipClick`s.
- **Wired `lossShow` on 17 casino pages** whose result model separates a loss from a push (PR 1 made `lossThud` opt-in for exactly this reason). The 5 pages with compound `humanWonMatch || result > 0` conditions were left alone — their loss semantics are ambiguous.
- **Removed the shared sound wiring from `useCommunityPokerGame`** (serves 4 pages).
- **Removed the temporary 3 s `winFanfare`/`lossThud` dedupe guard** now that the page-level handlers are gone.
- **Cleanup**: untracked the repo-root `public/sounds/` build output and gitignored it (it was partially tracked, so every build dirtied the tree), deleted dead `useHaptics.ts` + test, knip clean.

### Two behavior traps the sweep caught (worth a reviewer's eye)

1. **`errorBuzz` is only safe to delete on pages that render `ErrorAlert`.** 196 of 222 pages do, but FiftyOne and War show errors with their own inline retry link, so deleting their effect silenced failures entirely. Both keep a page-level buzz with a comment explaining why.
2. **The sound is now asynchronous.** Page handlers played on click; the central tap plays after the exec resolves. Tests needed `await waitFor`, and the ear pass should confirm move feedback still feels immediate.

## Test plan

- [x] Full frontend suite (1132 files / 15495 tests, 0 failures, 0 unrun — NavBar excluded locally per its known heap OOM) + build (tsc) + Biome green
- [x] `grep -rn "playSound('winFanfare'\|playSound('cardPlace'\|playSound('shuffle'" frontend/src` returns only manifest-approved KEEP sites
- [x] knip reports no new dead code (2 pre-existing findings unrelated to this branch)
- [ ] Prod-build ear pass: five games across casino/classic/solo/extra — one sound per action, one fanfare per win, no beep on an illegal move, rising arpeggio on a solitaire click-run
- [ ] CI green


yuta-yoshinaga:
## Review response — leftover empty blocks removed

Accurate catch, and thanks for the multiline scan: the sweep emptied the `if (...) { }` guards instead of deleting them, leaving the block, the `useRef` feeding it, and a comment describing behavior the code no longer performs. Biome does not flag empty block statements in this config, so `bun run check` was green and it would have shipped.

Fixed in the latest commit by deleting all four outright:

- `PrsiPage`: the error-buzz effect (ErrorAlert owns it) and the empty branch in `handlePlayWithSound` — the wrapper had nothing left to add, so callers now use `handlePlay` directly. Its **draw** sound stays, with the reason inline: the central tap maps only the `reset` command to shuffle, so this action would otherwise get the generic card sound.
- `BakersDozenPage`: both the `moveCount` precision effect and the error-buzz effect. The former is safe to drop because PR #4338's rejected-action gate keeps the central sound silent on illegal moves — the exact precision that effect existed to provide.
- Plus the imports tsc then flagged as unused.

Verification (the toolchain you couldn't run):
- `bun run build` (incl. tsc) ✅, `bun run check` ✅
- Full frontend suite: **1132 files / 15495 tests, 0 failures, 0 unrun** (NavBar excluded locally — it heap-OOMs the worker on this machine and silently skips ~200 other tests; CI runs it)
- Re-ran your scan repo-wide for `{\n\s+}` across `src/`: **zero** empty blocks remain.

PR #4341, opened 2026-07-25
chore(deps): update frontend and Go dependencies

yuta-yoshinaga:
## Summary

Updates frontend and Go dependencies. One commit per upgrade so a regression can be reverted in isolation.

| Commit | What |
|---|---|
| `chore(deps): update frontend dependencies within their major ranges` | react/react-dom 19.2.4→19.2.8, @tanstack/react-query 5.90→5.101, framer-motion 12.38→12.42, react-router-dom 7.13→7.18, mermaid 11.13→11.16, vite 7.3.1→7.3.6, vitest + coverage-v8 4.1.0→4.1.10, playwright 1.58→1.62, tailwindcss 4.2→4.3, happy-dom 20.8→20.11, typedoc 0.28.17→0.28.20, @types/* patches — **plus the `server.fs.allow` fix below** |
| `chore(deps): update biome 2.4.8 -> 2.5.5` | formatter reflows one test file; no new lint failures |
| `chore(deps): update i18next 25 -> 26 and react-i18next 16 -> 17` | react-i18next 17 requires i18next >= 26.2, so they move together |
| `chore(deps): update @testing-library/jest-dom 6 -> 7` | peer `@testing-library/dom >=10 <11` satisfied by the installed 10.4.1 |
| `chore(deps): update golang.org/x modules` | crypto 0.52→0.54, sys 0.45→0.47, term 0.43→0.45 |
| `chore(deps): declare jsdom explicitly` | it was an undeclared dependency that this PR's vitest bump knocked out — see below |

## The one real breakage: a *minor* Vite bump broke 222 test files

Vite 7.3.2+ denies serving files outside the project root. `src/constants/cuiManualTexts.ts` imports all 219 CUI manuals from the repo-root `docs/` tree via `?raw`, so every test that reaches a game page failed to collect with `Denied ID .../docs/manual/accordion.md?raw`.

The nasty part: **`bun run build` stayed green throughout**, because production inlines those imports at build time — only the dev server and Vitest hit the restriction. Fixed with an explicit `server.fs.allow: ['..']` plus a comment naming the failure mode.

## The second breakage: an undeclared dependency CI could see and local runs could not

13 test files select the jsdom environment via an `@vitest-environment jsdom` docblock, but `jsdom` was never in `package.json`. It was present only as a resolved entry for vitest's **optional** peer dependency, and the 4.1.0 → 4.1.10 bump re-resolved those optional peers and dropped it.

Local runs stayed green the whole time on a stale `node_modules` copy of jsdom 28.1.0. CI builds with `bun install --frozen-lockfile`, got a tree without it, and 13 workers failed to start with `ERR_MODULE_NOT_FOUND`. Now declared explicitly, and verified against a clean `--frozen-lockfile` install in a scratch directory rather than trusting the local tree again.

## Deliberately NOT updated

| Held back | Why |
|---|---|
| TypeScript 5.9 → 7.0 | the native compiler port; typedoc/biome/vitest support is an unknown, wants its own evaluation |
| Vite 7 → 8 | major, and it drags `@vitejs/plugin-react` with it |
| @vitejs/plugin-react 5 → 6 | peer-requires `vite ^8.0.0`, so it is coupled to the Vite 8 decision |
| @types/node 24 → 26 | the runtime here is Node 24; types ahead of the runtime let tsc accept APIs that do not exist |
| `go 1.25.8` directive | pinned for TinyGo per ADR |
| syumai/workers 0.32 → 0.33 | changes the TinyGo WASM output size, and the `extra` worker is already at the 1 MB free-tier ceiling |
| go-runewidth 0.0.3 → 0.0.27 | large jump in CUI column-width behavior; wants a CLI-rendering pass of its own |

## Test plan

- [x] Frontend: `bun run build` (incl. tsc) + `bun run check` green
- [x] Frontend suite: **1133 files / 15984 tests, all passing, zero unrun** (rebased onto #4340 so the worker-heap fix is present — without it NavBar OOMs and silently skips ~212 tests, which made the first two runs of this branch unreadable)
- [x] Go: `go build ./...`, full `go test -tags test ./...` (16 packages ok), `golangci-lint run ./...` → 0 issues
- [ ] CI green (incl. the TinyGo worker size check, which the Go bumps could in principle move)

_Correction (from review): an earlier version of this description and the Go commit message also listed `net` and `text`. `go get` printed "upgraded" for both, but neither is a real dependency of this module — they appear only as pre-2022 `/go.mod`-only checksum stubs with no `h1:` hash, and `go.mod`/`go.sum` show them unchanged. Only crypto, sys and term actually moved._


yuta-yoshinaga:
Verified and fixed — the nit was correct.

`golang.org/x/net` and `golang.org/x/text` are **not** dependencies of this module: they appear in `go.sum` only as pre-2022 `/go.mod`-only stubs with no `h1:` hash (`net v0.0.0-20190404...`, `text v0.3.0`), they are absent from `go.mod` entirely, and `git diff` on that commit shows zero change to those lines. `go get` printed `upgraded golang.org/x/net v0.54.0 => v0.57.0` for them, which is what I copied into the message — but that was the module graph's requirement, and `go mod tidy` left no resolved dependency behind.

Amended the commit body to list only crypto/sys/term, with a note explaining the `go get` output so the next person doesn't repeat my mistake, and corrected the PR description table.

Good catch on exactly the thing a bisectable dependency PR must get right.

PR #4343, opened 2026-07-26
perf(ci): parallelize and shard the two slowest jobs

yuta-yoshinaga:
## Measured first

Every PR waits on two ~20 minute jobs. Per-step timings from a recent `develop` run:

| Step | Time |
|---|---|
| Frontend Tests → run with coverage | **1205s (20m)** |
| E2E → the test run | **~20m** |
| Backend Tests → run with coverage | 267s |
| E2E → build frontend | 78s |
| E2E → build server binary | 39s |
| E2E → install Playwright browsers | 27s |

## E2E was running single-threaded for no reason

`playwright.config.ts` had `workers: process.env.CI ? 1 : undefined` with no explanation. Nothing requires it: the server keys game state by `sessionId` (`base_controller.go` → `SessionProvider.Acquire`) and the frontend mints a fresh one per page load (`gameApi.ts:244`, `crypto.randomUUID()`), so two specs cannot land on the same game state.

Raising it to 4 on a 20-spec sample: **1m24s → 38s**, 32/32 passing. Pinned to 4 rather than `undefined` so a larger runner does not aim one Chromium per core at a single backend.

## Both jobs sharded

Frontend 4 ways, E2E 3 ways (on top of the 4 workers). Each E2E shard boots its own server binary, so shards are independent.

**Why sharding and not "just drop coverage":** on a 16-core box the frontend suite still takes **10m43s without coverage** (85 min of CPU time). The cost is the 15,984 tests, not the v8 instrumentation — dropping coverage would surrender the 80% branch-coverage gate and buy almost nothing.

## Verified the splits are lossless before trusting them

A shard config that silently drops tests is worse than a slow one, so I checked the arithmetic rather than assuming:

- vitest: 3658 + 3746 + 4458 + 4122 = **15,984** — exactly the unsharded total
- Playwright: 102 + 102 + 102 = **306** — exactly the unsharded total
- One E2E shard end-to-end locally: **3m0s**, 102/102 passing

Expected wall clock: frontend ~5-6 min, E2E ~3-4 min, so the critical path drops from ~20 min to roughly the Backend Tests' 4.5 min.

## Test plan

- [x] Shard arithmetic verified lossless on both runners (above)
- [x] E2E shard 1/3 at 4 workers: 102 passed in 3m0s
- [x] 20-spec parallel sample: 32 passed, 2.2× faster
- [x] Workflow YAML lints
- [ ] CI green, and the wall-clock improvement confirmed on the real runners
- [ ] Codecov still reports merged frontend coverage across the 4 shard uploads

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,8 +67,18 @@ jobs:
         run: bun run check
 
   test-frontend:
-    name: Frontend Tests
+    # Sharded because this is the slowest job in the pipeline (~20 min
+    # unsharded): 15,984 tests over 1,133 files. Measured on a 16-core box the
+    # run still takes ~11 min WITHOUT coverage, so the cost is the test volume
+    # itself, not the coverage instrumentation — splitting the work is the only
+    # lever that moves it. Each shard uploads its own coverage; Codecov merges
+    # them server-side.
+    name: Frontend Tests (${{ matrix.shard }}/4)
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        shard: [1, 2, 3, 4]
     steps:
       - uses: actions/checkout@v6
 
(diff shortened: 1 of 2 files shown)


yuta-yoshinaga:
## Measured on the real runners — critical path 17.4m → 7.6m

| Job | Before | After | |
|---|---|---|---|
| E2E Tests | **1044s (17.4m)** | 250 / 308 / **370s** across 3 shards | −65% |
| Frontend Tests | **928s (15.5m)** | 226 / 233 / 288 / **454s** across 4 shards | −51% |
| Backend Tests | 289s | 298s | unchanged |
| Lints | ~20s | ~20s | unchanged |

The pipeline is gated by its slowest job, so the wait drops from **17.4 min to 7.6 min** — and Backend Tests (298s) is now within striking distance of becoming the new critical path.

Correctness held: all 7 sharded jobs green, and **codecov/patch passed**, confirming Codecov merges the four partial frontend uploads rather than reporting one shard's view.

Shard balance is uneven (Frontend 226-454s, E2E 250-370s) because both runners split by file count, not by runtime. Evening that out would buy another ~2 min on the frontend job, but it needs per-file timing data to do properly, so I left it rather than guess.

PR #4345, opened 2026-07-26
perf(test): make the NavBar active-link assertion O(n) instead of O(n²)

yuta-yoshinaga:
## The frontend job's floor was one file, and one loop inside it

After #4344 the six frontend shards land at 192, 196, 207, 207, 211 … and **405s**. That outlier is not scheduler imbalance — it is the shard holding `NavBar.test.tsx`, which takes **149s on its own with zero contention** (489 tests, 19× the next slowest file at 30s). No shard count can go below a single file's runtime, so more shards could never move it.

The cost turned out to be algorithmic:

```ts
for (const { path, labelKey } of gameRoutes) {        // 219 routes
  it(`marks ${labelKey} link as active when on ${path}`, () => {
    renderNavBar(path);
    expect(screen.getByText(labelFor(labelKey))).toHaveAttribute('aria-current', 'page');
    for (const other of gameRoutes) {                 // × 218 more lookups
      if (other.path !== path) {
        expect(screen.getByText(labelFor(other.labelKey))).not.toHaveAttribute('aria-current');
      }
    }
  });
}
```

219 × 218 ≈ **48,000 `getByText` scans**. Collecting the `aria-current` links in one pass gives the same guarantee in O(n):

```ts
const current = screen.getAllByRole('link').filter((link) => link.hasAttribute('aria-current'));
expect(current).toHaveLength(1);
expect(current[0]).toHaveAttribute('aria-current', 'page');
expect(current[0]).toHaveTextContent(labelFor(labelKey));
```

**149s → 78s, all 489 tests still passing.**

## Equivalence verified by mutation, not by reading

A faster test that catches less is a downgrade, so I broke the component on purpose and compared old vs new:

| Mutation in `NavBar.tsx` | Old test | New test |
|---|---|---|
| every game link gets `aria-current="page"` | 219 failed | **219 failed** |
| the wrong link gets `aria-current` (`===` → `!==`) | 219 failed | **219 failed** |

Identical detection, both times.

(A first attempt at this mutated lines 176/199 and *nothing* failed — those sit in the mobile favourites/recents branches, which the default render never reaches. The old test passed too, which is what flagged the experiment as invalid rather than the rewrite as weak. The real link list is at 227/263.)

## Expected effect

That shard should drop from ~405s toward the ~200s band the other five already occupy, taking the pipeline's critical path with it. Same file also OOM'd the vitest worker earlier today (#4340) — halving its DOM churn should ease that pressure too.

## Test plan

- [x] 489/489 passing, 149s → 78s locally
- [x] Mutation-tested for equivalence against the previous assertion (table above)
- [x] Biome clean
- [ ] CI green, and the slow shard measurably closer to the others

### DIFF
--- a/frontend/src/components/NavBar.test.tsx
+++ b/frontend/src/components/NavBar.test.tsx
@@ -54,10 +54,8 @@ describe('NavBar', () => {
 
     it(`marks ${labelKey} link as active when on ${path}`, () => {
       renderNavBar(path);
-      // Collect the links carrying aria-current in one pass and assert there is
-      // exactly one. Equivalent to checking every other route individually, but
-      // O(n) instead of O(n^2) — the per-route version ran 219 getByText queries
-      // per test across 219 tests (~48k DOM scans) and dominated the suite.
+      // Exactly one link may carry aria-current, which also proves no other
+      // route's link has it.
       const current = screen.getAllByRole('link').filter((link) => link.hasAttribute('aria-current'));
       expect(current).toHaveLength(1);
       expect(current[0]).toHaveAttribute('aria-current', 'page');
--- a/frontend/src/components/NavBar.test.tsx
+++ b/frontend/src/components/NavBar.test.tsx
@@ -54,12 +54,14 @@ describe('NavBar', () => {
 
     it(`marks ${labelKey} link as active when on ${path}`, () => {
       renderNavBar(path);
-      expect(screen.getByText(labelFor(labelKey))).toHaveAttribute('aria-current', 'page');
-      for (const other of gameRoutes) {
-        if (other.path !== path) {
-          expect(screen.getByText(labelFor(other.labelKey))).not.toHaveAttribute('aria-current');
-        }
-      }
+      // Collect the links carrying aria-current in one pass and assert there is
+      // exactly one. Equivalent to checking every other route individually, but
+      // O(n) instead of O(n^2) — the per-route version ran 219 getByText queries
+      // per test across 219 tests (~48k DOM scans) and dominated the suite.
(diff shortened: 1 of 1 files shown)


yuta-yoshinaga:
## Measured on CI: the slow shard came back to the pack

| | Before (#4344) | After |
|---|---|---|
| Frontend shards | **405** / 211 / 207 / 207 / 196 / 192 | **293** / 210 / 207 / 205 / 189 / 155 |
| E2E shards | 356 / 301 / 258 | 272 / 253 / 231 |
| Critical path | 405s | **293s** |

The outlier shard dropped 28% and is now within ~80s of the others instead of ~200s. E2E also improved without any change of its own — that is #4344's Playwright browser cache hitting for the first time (it could only populate on the run that introduced it).

Review nit taken: the inline comment carried the fix's history and measurements, which CLAUDE.md explicitly says belongs in the PR description because it rots in code. Trimmed to the one durable claim — *exactly one link may carry aria-current* — and re-verified (489/489, 78s).

Thanks for independently deriving the equivalence from  (brand link never sets it,  is outside , favourites/recents are mobile-only,  yields the category view). That is the same reasoning the mutation runs confirmed empirically, arrived at from the other direction.

yuta-yoshinaga:
(Correcting the previous comment — backticks got eaten by my shell, dropping the identifiers from the last paragraph.)

Thanks for independently deriving the equivalence from `NavBar.tsx`: the brand link never sets `aria-current`, `/discover` sets it but is outside `gameRoutes`, favourites/recents only render when `isMobile`, and `useGameRouteSearch('')` returns `filteredRoutes: null` so the category view is what renders. That is the same conclusion the mutation runs reached empirically, arrived at from the opposite direction — static reading vs. breaking the component and watching which tests notice.

PR #4349, opened 2026-07-26
perf(e2e): skip the CPU replay animation and race the Sevens loop

yuta-yoshinaga:
## The remaining E2E cost was the app's animation, not the tests

I went after `sevens.spec.ts` expecting the #4342 shape — a 300-iteration loop probing controls one at a time. Fixing that gave **53.4s → 45.9s**, far less than expected, so I instrumented the loop instead of accepting it:

```
turn 1  2878ms
turn 2  2862ms
turn 3  2869ms
turn 4  2866ms
...
```

A flat ~2.87s per turn regardless of what the turn did. That is three CPU actions at `REPLAY_DELAY_MS = 800ms` — the replay animation, which no amount of test-side restructuring can remove.

## The lever was already in the app

`useReplaySpeed` is a user setting (`normal` / `fast` / `instant`) stored in localStorage, read fresh on every replay step, and `instant` means multiplier **0**. `navigateTo` already seeds localStorage to suppress the tutorial dialog, so one more line covers every spec that drives CPU turns:

```ts
localStorage.setItem('cpuReplaySpeed', 'instant');
```

| | Before | After |
|---|---|---|
| `sevens.spec.ts` | 53.4s | **4.1s** |
| Full suite (4 workers, local) | 10.1m | **6.0m** |

306/306 still passing, which is the check that matters here: nothing in the suite was asserting on the animation being visible. Specs assert on the state a replay lands in, not on its pacing.

## Also included: the Sevens loop fix

Kept because it is correct on its own merits — the loop probed the end-state with a 1s timeout every turn even mid-game, the same budget problem as #4342. It just was not the dominant cost, and finding that out is what led to the replay setting.

## Effect on CI

E2E shards are ~105s setup + test time; this attacks the test-time half. With 5 shards from #4348 they ran 171-227s, so this should pull them toward the setup floor.

## Test plan

- [x] `sevens.spec.ts`: 53.4s → 4.1s
- [x] Full E2E suite: **306/306 passing**, 10.1m → 6.0m locally at 4 workers
- [x] Biome clean
- [ ] CI green, E2E shards closer to their ~105s setup floor

### DIFF
--- a/frontend/e2e/helpers.ts
+++ b/frontend/e2e/helpers.ts
@@ -35,6 +35,10 @@ export async function navigateTo(page: Page, path: string) {
   // Suppress the first-visit tutorial suggestion dialog to avoid blocking game interaction
   await page.addInitScript(() => {
     localStorage.setItem('tutorial_no_suggest', 'true');
+    // Skip the CPU replay animation. Specs assert on state, not on the
+    // pacing of the animation, and at the default 'normal' speed every
+    // CPU action costs 800ms — Sevens alone spent ~2.9s per turn waiting.
+    localStorage.setItem('cpuReplaySpeed', 'instant');
   });
   await page.goto(`/#${path}`);
   await waitForLoaded(page);
--- a/frontend/e2e/sevens.spec.ts
+++ b/frontend/e2e/sevens.spec.ts
@@ -1,5 +1,5 @@
 import { expect, test } from '@playwright/test';
-import { isVisibleWithin, navigateTo, TIMEOUT_ACTION, TIMEOUT_QUICK, waitForLoaded } from './helpers';
+import { navigateTo, TIMEOUT_GAME_LOOP, waitForLoaded } from './helpers';
 
 test.describe('Sevens E2E', () => {
   test('plays a full game: reset → play or pass each turn → end → reset', async ({ page }) => {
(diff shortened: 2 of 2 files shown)


yuta-yoshinaga:
## Correction: this helps local runs, not the CI critical path

I claimed the suite goes 10.1m → 6.0m. That is true locally and I should have checked what it does to CI before implying it moves the pipeline. Measured:

| | test-time sum across 5 shards | setup avg |
|---|---|---|
| Before (#4348) | 457s — [104, 103, 91, 85, 74] | 108s |
| After | **444s** — [121, 97, 96, 69, 61] | 107s |

**-3%, inside the noise.** One shard dropped 85→61s, another rose 104→121s; runner variance here is ±20s, so sevens' ~49s saving lands in a single shard and gets swamped.

Two reasons the local win does not transfer:

1. **E2E is setup-bound now.** ~107s of every shard is frontend build + browsers + server binary, which no test-side change touches. Test time is already the smaller half.
2. **E2E is not the critical path.** That is the frontend's 273s shard (6/8). This PR cannot move the pipeline no matter how fast the specs get.

## Still worth merging

- The mechanism is verified, twice: `sevens.spec.ts` **53.4s → 4.1s**. `addInitScript` seeding localStorage behaves identically on any runner.
- **306/306 passing** with the animation skipped, which is the real assurance: no spec was depending on replay pacing being visible.
- Local `bun run e2e` goes 10.1m → 6.0m, which is where a developer actually feels it.
- The Sevens loop fix is correct on its own merits (a 1s end-state probe every turn, the #4342 pattern).

Framing it as a CI win was wrong; framing it as a dev-loop win with a verified 92% cut on the worst spec is accurate.

PR #4350, opened 2026-07-26
perf(test): assert every NavBar route from one render

yuta-yoshinaga:
## One file holds the whole pipeline

CI's slowest frontend shard (6/8) runs 273s while the other seven finish in 154-168s. Inside it:

| file | duration |
|---|---|
| **`NavBar.test.tsx`** | **110.8s** |
| `OmahaPage.test.tsx` | 14.3s |
| `KlondikePage.test.tsx` | 11.8s |
| …138 more | — |

42% of the shard in one file, 8× the next. Sharding cannot help — a file runs in one worker.

## The cost is the render count, not any slow assertion

Per-test timings are unremarkable (~140ms average). The problem is how many there are:

| group | tests | time |
|---|---|---|
| `marks X link as active` | 219 | 33.1s |
| `renders X link` | 219 | 27.9s |
| everything else | 51 | 14.2s |

Two loops over 219 routes re-render the entire NavBar **438 times**.

## `renders X link` was already covered

Three assertions overlapped:

- **[A]** `renders ${labelKey} link` × 219 — "this label is in the DOM"
- **[B]** `renders navigation links for all game routes` × 1 — only counted links
- **[C]** `links point to correct hrefs` × 1 (570ms after #4347) — asserts **every** route's label *and* href from one render

[C] strictly contains [A]. So [A] is deleted and [B] is tightened to assert every label is rendered rather than just counting.

**76s → 43s, 489 → 270 tests.**

## Equivalence: the single-route mutation is the real test

Deleting 219 tests in favour of one bulk assertion is exactly the change that can quietly lose resolution — a bulk check might catch "everything is broken" but miss "one route is broken". So that is the mutation I ran:

| Mutation in `NavBar.tsx` | Old suite | New suite |
|---|---|---|
| **exactly one route's label broken** (`nav.blackjack` → `'BROKEN'`) | 3 failed | **2 failed** |
| every route's label blanked | 440 failed | 222 failed |

The single-route case is caught either way. Failure counts differ only because the deleted tests no longer exist to fail.

## Expected effect

That shard should fall from 273s toward the 154-168s band the other seven occupy, taking the critical path with it — and without adding runners, which matters because a develop push already queues one shard at the 20-job concurrency limit.

## Test plan

- [x] 270/270 passing; file 76s → 43s
- [x] Mutation-tested including the single-route case (table above)
- [x] Biome clean
- [ ] CI green, shard 6/8 measurably closer to the pack

### DIFF
--- a/frontend/src/components/NavBar.test.tsx
+++ b/frontend/src/components/NavBar.test.tsx
@@ -37,6 +37,13 @@ describe('NavBar', () => {
     const links = screen.getAllByRole('link');
     // game links + brand link
     expect(links.length).toBeGreaterThanOrEqual(gameRoutes.length);
+    // Every route's label must be rendered. Asserting this from one render
+    // replaces a per-route test that re-rendered the whole NavBar 219 times.
+    const rendered = new Set(links.map((link) => link.textContent ?? ''));
+    for (const { labelKey } of gameRoutes) {
+      const label = labelFor(labelKey);
+      expect([...rendered].some((text) => text.includes(label))).toBe(true);
+    }
   });
 
   it('renders category labels for all categories', () => {
@@ -47,11 +54,6 @@ describe('NavBar', () => {
   });
 
   for (const { path, labelKey } of gameRoutes) {
-    it(`renders ${labelKey} link`, () => {
-      renderNavBar();
-      expect(screen.getByText(labelFor(labelKey))).toBeInTheDocument();
-    });
-
     it(`marks ${labelKey} link as active when on ${path}`, () => {
       renderNavBar(path);
       // Exactly one link may carry aria-current, which also proves no other
--- a/frontend/src/components/NavBar.test.tsx
+++ b/frontend/src/components/NavBar.test.tsx
@@ -37,13 +37,6 @@ describe('NavBar', () => {
     const links = screen.getAllByRole('link');
     // game links + brand link
     expect(links.length).toBeGreaterThanOrEqual(gameRoutes.length);
-    // Every route's label must be rendered. Asserting this from one render
-    // replaces a per-route test that re-rendered the whole NavBar 219 times.
-    const rendered = new Set(links.map((link) => link.textContent ?? ''));
-    for (const { labelKey } of gameRoutes) {
-      const label = labelFor(labelKey);
-      expect([...rendered].some((text) => text.includes(label))).toBe(true);
-    }
   });
 
   it('renders category labels for all categories', () => {


yuta-yoshinaga:
Confirmed and fixed — the blind spot is real, and worse than the review's count.

`nav.poker` is `'ポーカー'`, which appears inside **16** other `nav.*` labels (ビデオポーカー, ジョーカーポーカー, インディアンポーカー, フォーカードポーカー, オアシスポーカー, ロシアンポーカー, アイリッシュポーカー, 中国ポーカー, ポーカースクエアズ, パイガオ, スリーカード, キャリビアンスタッド, テキサスホールデムボーナス, パイナップル, オープンフェイスチャイニーズ, and the category label).

Reproduced exactly as predicted. With `/poker`'s label replaced by `'BROKEN'`:

```
× marks nav.poker link as active when on /poker
× links point to correct hrefs
× clicking a game link closes the menu
```

and running **only** the loop I added:

```
Tests  1 passed | 269 skipped (270)
```

So it never caught the case it was added for. The detection was coming from `links point to correct hrefs`, which groups link text **by href before comparing** and therefore has no collision exposure — exactly as you said.

Removed it. The file stays at 44s, which is its own proof the loop was doing no work worth paying for.

Equivalence re-verified on both a colliding and a non-colliding label this time:

| Mutation | Old suite | New suite |
|---|---|---|
| `nav.poker` label broken (16 colliding labels) | 4 failed | **3 failed** |
| `nav.blackjack` label broken (no collision) | 3 failed | **2 failed** |

The lesson I'll carry: I picked `nav.blackjack` for the mutation because it was the first route in the list, and a mutation target chosen for convenience tests the convenient case. Choosing the *adversarial* input — a label that is a substring of others — is what makes the check mean something.

yuta-yoshinaga:
## Measured: the outlier shard rejoined the pack

| | Before | After |
|---|---|---|
| Frontend (8 shards) | **271** / 168 / 160 / 159 / 158 / 157 / 154 / 154 | **210** / 164 / 163 / 162 / 160 / 158 / 155 / 151 |
| E2E (5 shards) | 227 / 225 / 199 / 196 / 171 | **230** / 205 / 187 / 182 / 168 |
| Backend | 201s | 205s |
| **Critical path** | **273s** | **230s** |

The frontend shard that held the pipeline dropped 271s → 210s, and the spread narrowed from 117s to 59s. Frontend is no longer the critical path — E2E's 230s is, and that job is ~107s setup, so it is close to its floor.

Note the file is 44s either way with the flawed loop removed, so all of this came from deleting the 219 duplicated per-route tests, not from the assertion I had added.

PR #4352, opened 2026-07-26
perf(ci): run CodeQL after the merge, not on every PR

yuta-yoshinaga:
## I optimized the wrong workflow for most of this session

The CI workflow went 1044s → 230s across #4343-#4351. Meanwhile CodeQL sat next to it, unmeasured, at **459s**. Measured across every workflow a PR triggers:

| Workflow | Slowest job |
|---|---|
| **CodeQL — Analyze (go, autobuild)** | **459s** |
| CI — E2E shard | 230s |
| Cloudflare workers build (tinygo ×4) | 225s |
| CodeQL — Analyze (javascript-typescript) | 218s |

CodeQL was the actual wait the whole time. Credit for the pointer goes to the review comment asking about it — I had been reading `gh pr checks` totals without ever opening those job logs.

## There is nothing to trim inside the Go job

459s = **250s extraction** + **120s writing the database** + **51s queries**. Module downloads are already cached (zero `go: downloading` lines in the log). The extraction cost is compiling 219 games; `autobuild` has to build the project to trace it.

**Not** switching to `build-mode: none`, which is the obvious lever and what the JavaScript job already uses. CodeQL's own help text, quoted from this run's log:

> buildless extraction will generally yield **less accurate analysis results**, and should only be used in cases where it is not possible to build the code

Trading detection accuracy for speed in the security scan is the wrong trade, and it would be an odd one to make silently.

## So change the trigger, not the tool

`develop` is an integration branch; `master` is what triggers releases. Scanning on the merge commit still catches a finding long before it can ship, and the weekly cron still covers query-pack updates against unchanged code. Alerts land in the Security tab exactly as before.

**What is genuinely lost:** the pre-merge signal. A vulnerability introduced by a PR is now reported minutes after the merge instead of before it. On a repo where `develop` is unprotected and merges are already made without waiting for CodeQL (I did exactly that twice today), that signal was not being used as a gate anyway — but it is a real reduction, not a free win, and worth being explicit about.

Reverting is a three-line change if you would rather have the pre-merge scan back.

## Effect

PR wait drops from ~459s to ~230s (the E2E shard), with the Cloudflare tinygo build at 225s right behind it.

## Test plan

- [x] Workflow YAML lints; triggers are now `push: [develop]` + weekly `schedule`
- [ ] Confirm CodeQL runs on the merge commit and its alerts still appear in the Security tab
- [ ] Confirm no CodeQL checks appear on the next PR

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -11,11 +11,14 @@
 #
 name: "CodeQL"
 
+# Runs after the merge, not on the PR. The Go job takes ~460s and the
+# JavaScript one ~220s, which made CodeQL — not the CI workflow (~230s) — the
+# real wait on every PR. develop is an integration branch, not a release
+# branch (master is), so scanning right after the merge still catches a finding
+# before it can ship, and the weekly run covers query-pack updates against
+# code that has not changed. Findings land in the Security tab either way.
 on:
   push:
-    branches: [ develop ]
-  pull_request:
-    # The branches below must be a subset of the branches above
     branches: [ develop ]
   schedule:
     - cron: '37 4 * * 0'
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -107,7 +107,7 @@ See [`docs/new-game-checklist.md`](docs/new-game-checklist.md) for the full chec
 
 ## Git Workflow
 
-- **`develop`**: Default branch; target for all PRs. CodeQL analysis and `golangci-lint` run on push/PR.
+- **`develop`**: Default branch; target for all PRs. `golangci-lint` and the CI suite run on push/PR; CodeQL runs on push to `develop` (i.e. after the merge) and weekly — see [ADR-0034](docs/adr/0034-codeql-post-merge.md).
 - **`master`**: Triggers automatic version bump, git tag, and GitHub Release.
 - **PR Summary**: When creating a PR, if there is an associated issue, the PR description must explicitly close the issue (e.g., `Closes #123`).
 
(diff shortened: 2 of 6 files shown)


yuta-yoshinaga:
All three are correct, and the ADR point is the one I should have caught myself — this repo's rules spell it out and I read them at the start of the session.

Fixed in the follow-up commit:

**1. `CLAUDE.md:110`** — now reads that `golangci-lint` and CI run on push/PR while CodeQL runs on push to `develop` (after the merge) and weekly, linking ADR-0034.

**2. `CONTRIBUTING.md:76`** — the pre-merge checklist said "CI（lint, test, CodeQL）がすべて通ることを確認してください", i.e. it instructed contributors to wait for a check that will never appear. Now lists lint/test/E2E, plus an explicit line saying CodeQL runs post-merge with results in the Security tab.

**3. ADR** — ADR-0019 is only *partially* contradicted: its golangci-lint, CI and auto-tag decisions all still hold, and only the CodeQL trigger changed. Marking the whole thing `Superseded` would have thrown away three still-valid decisions, so I annotated that one bullet in place and wrote **ADR-0034** for the new decision. It records:

- the measurement that motivated it (CodeQL Go 459s vs the CI workflow's 230s — CodeQL was the real PR wait all along)
- the alternatives, including **why `build-mode: none` was rejected** despite being the obvious speedup, quoting CodeQL's own log text about buildless extraction being less accurate
- the consequence being accepted, stated as a loss and not a win: pre-merge detection becomes post-merge detection
- the three-line revert, plus the note that making it a real gate would also need branch protection on `develop` (currently unprotected)

Index updated, with ADR-0019's row flagged as partially replaced so the trail is followable from either direction.

PR #4433, opened 2026-07-27
refactor(frontend): split card.ts into per-game type modules behind a barrel

yuta-yoshinaga:
## Summary

`card.ts` held **894 declarations in 11,212 lines** and was the single most-changed file in the repo, because all 219 games route through it. **880 of those declarations belong to exactly one game.**

It is now a **217-line barrel** over:
- `src/types/common.ts` — the 12 genuinely shared declarations
- `src/types/games/<game>.ts` — 210 files

**No import anywhere changes.** 1,383 files import from `types/card`; the barrel re-exports everything, so every one still resolves.

> **Scope: `card.ts` only.** `gameApi.ts` is deliberately left for its own PR, as the issue instructs — it's 5,409 lines of *runtime* code, where the tree-shaking question is real rather than theoretical. Commit says `Refs`, not `Closes`.

## Dependency graph, verified before moving anything

The issue warns that cross-game references need finding first. Measured: **11 pairs, acyclic**, in two families:

| family | depends on |
|---|---|
| `omaha`, `shortdeck`, `pineapple` (and `irishpoker` → `pineapple`) | `holdem` |
| `yukon`, `russiansolitaire`, `cruel`, `scorpion`, `wasp`, `easthaven` | `klondike` |

Those modules import from `./holdem` / `./klondike` directly. I ran a cycle check rather than eyeballing it.

Also resolved a naming mismatch: the `PigsTail*` declarations belong to the game registered as `pigtail`, so a plain prefix match left them unowned.

## The bundle is unchanged — and I nearly misreported this

My first comparison used chunk sizes I'd recorded during #4368 earlier in this batch. `QueryClientProvider` appeared to grow **+4.7 kB gzip** — precisely the tree-shaking regression the issue warns about. But #4369 and #4365 had merged in between, both of which add code.

Building `develop` for a real baseline:

```
BEFORE total gzip: 3648.68 kB
AFTER  total gzip: 3648.43 kB   (-0.25)
size-distribution mismatches across all 873 asset entries: 0
```

Which is exactly what a types-only change should do — types are erased at compile time, and `card.ts` had exactly **one** runtime export (`isMaskedCard`, now in `common.ts`). The apparent regression was entirely the intervening merges.

## Test plan

- [x] `bun run typecheck` — **clean on the first attempt**, which is the decisive check here: it proves all 1,383 dependent imports resolve and all 894 declarations landed with correct cross-module imports
- [x] `bun run test` — 15,841 passed
- [x] `bun run check` — clean, 0 warnings, all six guards green
- [x] `bun run build` — clean
- [x] **Bundle measured against a real `develop` baseline**, not against remembered numbers
- [x] `bun run e2e` before pushing — 305/307; both failures (`poker`, `crazyeights`) are the documented rotating flakes, and `poker` passes alone in 5.8s / 6.2s. A types-only change with a size-identical bundle has no mechanism to alter runtime behaviour.

## Follow-up

`gameApi.ts` (5,409 lines, 165 API objects + 7 factories) still needs the same treatment. Unlike this PR it ships runtime code, so it needs its own before/after bundle measurement — and per this repo's notes, the import *closure* matters more than the entry chunk.

Refs #4366


yuta-yoshinaga:
## Review fix

**Fixed the `common.ts` placement.** You were right that it broke the file's own stated rule. Verified the usage before moving: `CpuAction` and `DrawHistoryEntry` are consumed only by `oldmaid.ts`, `useOldMaidGame.ts`, `OldMaidPage.tsx` and `OldMaidDrawHistory.tsx` — nowhere else. Both moved into `types/games/oldmaid.ts`, which no longer imports them from `../common`.

Your latent-collision point is the stronger reason: a bare `CpuAction` in the shared namespace would force any other game wanting that name to work around it, which is precisely the footgun the `<Game>Xxx` prefixes avoid everywhere else in the split.

`common.ts` is now **10 declarations**, all genuinely cross-game:

```
CardDesign  BaseGameResponse  Card  MaskedCard  isMaskedCard
BettingProfileBracketData  BettingHumanProfileData  BettingMetaAI
ActionLogEntry  ActionLogResponse
```

## On the count — you're right, and here's the reconciliation

**895 exported declarations**, not 894. The difference is what gets counted:

- **894 type declarations** — 824 `interface` + 70 `type`, which is the figure the issue quotes
- **+1 runtime export** — `isMaskedCard`, the only value in the whole file

= **895 total**. After this fix: `common.ts` (10) + `games/*.ts` (885) = **895**. Nothing dropped, nothing duplicated. My "894" came from grepping only `interface|type`, so both numbers were describing real things — I just wasn't precise about which.

Thanks also for independently checking the things that would actually hurt: no name collisions across the 210 modules (the `export *` footgun), no per-game file re-importing `../card` (which would cycle through the barrel), and every new import being `import type` for `verbatimModuleSyntax`/`erasableSyntaxOnly`. Those are the failure modes I'd have wanted a second pair of eyes on.

## Verification

- [x] `bun run typecheck` — clean
- [x] `bun run check` — clean, 0 warnings, all six guards green
- [x] Old Maid suite (6 files, 178 tests) — passed
- [x] `bun run build` — clean
- [x] Bundle unaffected: both moved declarations are `interface`s, erased at compile time

PR #4438, opened 2026-07-27
test(frontend): assert the keyboard shortcuts actually run, on 16 solitaire pages

yuta-yoshinaga:
## Summary

Nothing verified that pressing a bound key does anything. The bindings were type-checked, their labels guarded (#4369's `kbd-labels` check), and the panel listing them tested — but **a wrong `key`, or a wrong `enabled` condition, would have failed no test.**

Adds **87 tests** across the 16 solitaire pages that had bindings and no `keyDown` coverage:

- one per key asserting the dispatch
- one asserting give-up asks for confirmation instead of firing (irreversible, per #2099)
- one asserting every key is inert after the game ends

## The tests bite — verified by mutation

Changing Yukon's `key: 'h'` to `'H'` fails `pressing h dispatches hint`; restoring it passes. That is precisely the defect class the issue describes, so the tests are worth their weight rather than just green.

## Correcting the issue's premise — which I wrote

#4429 claims keyDown coverage is "effectively zero". **42 of the 71 action-nav pages already had it.** The real gap is **27 pages / 130 bindings**, and this PR covers the 16 uniform solitaire ones. The remaining 11 are casino/poker pages (`IndianPoker` 9 bindings, `UltimateTexasHoldem` 8, `MississippiStud`/`OasisPoker`/`TexasHoldemBonus` 6 each, …) whose commands are less uniform; they belong in a second pass.

## Two pages needed individual handling

- **`Crescent`** — `handleRedeal` dispatches `'redeal'`, not the `'deal'` its label implied.
- **`Spiderette`** — `d` is gated by `handleDealGuarded`, which refuses to deal while a tableau column is empty and stock remains. Its fixture has four empty columns with stock 24, so the key is *correctly* inert; the test asserts the guard.

## Where I went wrong getting there

I first attributed Spiderette's failure to **Easthaven**, by matching two assertion errors to files via their mock variable name without checking which was which. I then "fixed" Easthaven by replacing a correct, passing test with a guard assertion its fixture never triggers. **Easthaven had never failed.**

My fixture-inspection regex also mis-reported Spiderette as having *no* empty column, because its tableau is built through a `makeTableau()` call the regex didn't look inside. Both pages are now back to what the evidence supports — Easthaven dispatches, Spiderette is guarded — and the empirical test result, not my parsing, was the reliable signal throughout.

## Test plan

- [x] `bun run test` — **15,929 passed (+87)**
- [x] `bun run typecheck` — clean
- [x] `bun run check` — clean, all six guards green
- [x] Mutation-tested one binding to confirm the tests fail when the wiring breaks
- [x] Test-only change: `git status` confirms no non-`.test.tsx` file is touched
- [x] E2E not run: this adds no production code (verified above), so there is nothing for it to exercise

Refs #4429 (16 of 27 pages; the 11 casino/poker pages remain)

### DIFF
--- a/frontend/src/hooks/useDoubtGame.ts
+++ b/frontend/src/hooks/useDoubtGame.ts
@@ -66,6 +66,23 @@ export function useDoubtGame() {
     if (state) cpuDoubtersRef.current = state.cpuDoubters;
   }, [state]);
 
+  // Clear the countdown interval on unmount. It was only ever cleared by
+  // stopCountdown or by its own tick reaching zero, so unmounting mid-countdown
+  // left it running: the next tick called setCountdown on a gone component, and
+  // in a test environment that is `ReferenceError: window is not defined` thrown
+  // from React's dispatchSetState — an unhandled error that fails the whole
+  // vitest run even when every test passed. Surfaced on CI by #4429's added
+  // tests shifting shard timing; latent before that.
+  useEffect(
+    () => () => {
+      if (countdownRef.current !== null) {
+        clearInterval(countdownRef.current);
+        countdownRef.current = null;
+      }
+    },
+    [],
+  );
+
   const stopCountdown = useCallback(() => {
     autoSkipRef.current = false;
     if (countdownRef.current !== null) {
--- a/frontend/src/pages/AcesUpPage.test.tsx
+++ b/frontend/src/pages/AcesUpPage.test.tsx
@@ -307,3 +307,46 @@ describe('AcesUpPage', () => {
     expect(actionLogApi).toBeDefined();
   });
 });
+
+// Keyboard shortcuts are bound by useActionKeyboardNav and advertised by
+// ActionShortcutsPanel, but nothing asserted that pressing a key actually runs
+// its action — a wrong `key` or a wrong `enabled` condition would have failed no
+// test. See issue #4429.
+describe('AcesUpPage keyboard shortcuts', () => {
+  it.each([
+    ['d', 'draw'],
+    ['h', 'hint'],
+    ['z', 'undo'],
+  ])('pressing %s dispatches %s', async (key, command) => {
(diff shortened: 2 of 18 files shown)


yuta-yoshinaga:
## CI failure was a real bug — fixed rather than rerun

Shard 6 failed **with every test passing**:

```
Test Files  142 passed (142)
Tests       2538 passed (2538)
Errors      1 error
ReferenceError: window is not defined
  at Timeout._onTimeout src/hooks/useDoubtGame.ts:83
```

`useDoubtGame`'s countdown interval was only ever cleared by `stopCountdown` or by its own tick reaching zero, so **unmounting mid-countdown left it running**. The next tick called `setCountdown` on a component that no longer exists, and in a torn-down test environment React's `dispatchSetState` reaches for `window` — an unhandled error that fails the whole vitest run even though nothing asserted wrongly.

**The leak is pre-existing and latent.** `DoubtPage` is not one of the 16 pages this PR touches. What changed is timing: the 87 added tests shifted the shard enough for a pending tick to land after teardown on a CI runner.

**Why I didn't just rerun.** I reproduced CI's exact invocation locally — `bun run test:coverage -- --shard=6/8`, same **142 files / 2538 tests**, so shard composition matches — and it passes here. That makes it timing-dependent, not composition-dependent, so a rerun would have been a coin flip that left a landmine for the next person whose tests shifted the clock.

Fixed with an unmount cleanup, plus a test that pins it. **Mutation-checked:** deleting the cleanup fails `clears the countdown interval when the page unmounts`; restoring it passes.

`bun run test` is now **15,930 passed**.

### Scope note

This adds a production change to what was a test-only PR. I judged that worth it: the failure is deterministic in the sense that it will recur for whoever next perturbs shard timing, and the fix is three lines with a test. Happy to split it out if you'd rather keep this PR test-only.

PR #4452, opened 2026-07-27
test(frontend): make the remaining 76 "not called" assertions real, and guard the pattern

### LINKED ISSUE #4451: test(frontend): 75 more vacuous assertions in hook tests — #4443's detector could not see the act() form
## Summary

75 `expect(<apiMock>).not.toHaveBeenCalled()` assertions across 52 hook test files pass unconditionally. Same root cause as #4439, **missed by that sweep** because of how its detector was written.

## Why #4439 missed them

#4439 found sites by looking for a `fireEvent` or `userEvent` call preceding the assertion with no `await` between. That is the page-test idiom. Hook tests drive the hook through `act(...)` instead:

```ts
act(() => result.current.handleDiscard());
expect(mockExec).not.toHaveBeenCalled();   // vacuous — react-query defers to a microtask
```

So `act(`-driven assertions were invisible to the detector, and #4443's "0 remaining" claim was true only of the pattern it searched for. Adding `\bact\(` to the same detector surfaces these 75.

## How this was found

While adding tests for four untested hooks (#4449 / #4450) I wrote exactly this vacuous shape myself, and mutation testing caught it: two deliberately broken selection guards left all 11 tests green. Re-running the detector with `act(` included then showed how widespread the pattern is.

Worth stating plainly: the lesson is not "remember to await" — it is that a detector's assumptions bound what it can find, so a clean sweep proves less than it appears to. #4443 reported zero remaining because it could not see this family at all.

## Fix

Same as #4443 — `await flushPendingDispatch()` before the assertion (`src/test/flushPendingDispatch.ts`, already fake-timer safe).

## Expect some to turn red, and treat that as the point

An assertion that has never actually run may be wrong. A red one means the guard it was written to verify is not holding — a production question, not a test fix. #4443 turned up a real 3-page bug this way (clicking a selected card dispatched a move onto its own column instead of deselecting). Batch these ~10 files at a time so any red is easy to attribute.

## Sites

| file | assertions |
|---|---|
| `hooks/useBakersDozenGame.test.ts` | 1 |
| `hooks/useBeleagueredCastleGame.test.ts` | 1 |
| `hooks/useBeziqueGame.test.ts` | 1 |
| `hooks/useBridgeGame.test.ts` | 1 |
| `hooks/useCalabresellaGame.test.ts` | 2 |
| `hooks/useCassinoGame.test.tsx` | 3 |
| `hooks/useCourtPieceGame.test.ts` | 1 |
| `hooks/useCrazyEightsGame.test.ts` | 2 |
| `hooks/useEcarteGame.test.ts` | 1 |
| `hooks/useEscobaGame.test.ts` | 1 |
| `hooks/useFlowerGardenGame.test.ts` | 1 |
| `hooks/useFortyFivesGame.test.ts` | 1 |
| `hooks/useFreeCellGame.test.ts` | 1 |
| `hooks/useGaigelGame.test.ts` | 1 |
| `hooks/useGinRummyGame.test.ts` | 4 |
| `hooks/useGoFishGame.test.ts` | 2 |
| `hooks/useHeartsGame.test.ts` | 2 |
| `hooks/useIndianRummyGame.test.ts` | 3 |
| `hooks/useJassGame.test.ts` | 1 |
| `hooks/useKingAlbertGame.test.ts` | 1 |
| `hooks/useKingGame.test.ts` | 1 |
| `hooks/useKlaverjasGame.test.ts` | 1 |
| `hooks/useKlondikeGame.test.ts` | 1 |
| `hooks/useKnockoutWhistGame.test.ts` | 1 |
| `hooks/useMacauGame.test.ts` | 1 |
| `hooks/useMachiavelliGame.test.ts` | 2 |
| `hooks/useManilleGame.test.ts` | 1 |
| `hooks/useMariasGame.test.ts` | 1 |
| `hooks/useNapGame.test.ts` | 1 |
| `hooks/useOmbreGame.test.ts` | 1 |
| `hooks/usePageOneGame.test.ts` | 1 |
| `hooks/usePanGame.test.ts` | 3 |
| `hooks/usePokerGame.test.ts` | 1 |
| `hooks/usePreferenceGame.test.ts` | 1 |
| `hooks/usePrsiGame.test.ts` | 2 |
| `hooks/useSambaGame.test.ts` | 3 |
| `hooks/useScoponeGame.test.ts` | 1 |
| `hooks/useSeahavenTowersGame.test.ts` | 1 |
| `hooks/useSedmaGame.test.ts` | 1 |
| `hooks/useSkatGame.test.ts` | 3 |
| `hooks/useSoloWhistGame.test.ts` | 1 |
| `hooks/useSpiderGame.test.ts` | 1 |
| `hooks/useSpoilFiveGame.test.ts` | 1 |
| `hooks/useStreetsAndAlleysGame.test.ts` | 1 |
| `hooks/useSuecaGame.test.ts` | 1 |
| `hooks/useTonkGame.test.ts` | 2 |
| `hooks/useTrickGameBase.test.ts` | 2 |
| `hooks/useTuteGame.test.ts` | 1 |
| `hooks/useTwentyNineGame.test.ts` | 1 |
| `hooks/useTysiacGame.test.ts` | 2 |
| `hooks/useUltiGame.test.ts` | 2 |
| `hooks/useWattenGame.test.ts` | 1 |

**Total: 75 assertions across 52 files.**

## Also worth doing

Add `\bact\(` to whatever detection is used, and consider whether this belongs in `scripts/check-design-tokens.mjs` as a guard so the pattern cannot come back — noting that the unmount-safety guard added in #4448 had to be rewritten three times before it stopped reporting safety it had not checked, so a test-file guard needs the same scepticism and a negative control.

Refs #4439, #4443, #4450

yuta-yoshinaga:
Closes #4451.

`expect(<apiMock>).not.toHaveBeenCalled()` placed straight after an interaction passes **whether or not the action fired** — `exec` dispatches through react-query's `mutateAsync`, so the assertion runs before the call can appear. #4443 fixed 148 of these. These 76 survived it.

## Why #4443 missed them

Its detector keyed on `fireEvent`/`userEvent` preceding the assertion — the *page-test* idiom. Hook tests drive the hook through `act(...)`:

```ts
act(() => result.current.handleDiscard());
expect(mockExec).not.toHaveBeenCalled();   // vacuous
```

So that whole family was invisible, and #4443's "0 remaining" was true only of the pattern it searched for. Widening again — to raw `.click()` — then turned up one more the `act`-aware pass had called clean (`EasthavenPage`'s deal guard).

**Three widenings, three times more sites.** The generalisable point: a clean scan means "clean under this pattern", not "clean".

## No production bugs this time — and that is a result, not a disappointment

#4443's sweep uncovered a real 3-page bug, so each failure here was worth checking. All 76 guards genuinely hold. What matters is that the assertions now *bite* rather than merely pass:

| guard disabled | before | after |
|---|---|---|
| Ulti's two-card discard | 9 passed ❌ | `× handleDiscard does nothing without exactly two selected cards` ✅ |
| Watten's one-card play | 9 passed ❌ | `× handlePlay does nothing without exactly one selected card` ✅ |
| Easthaven's deal guard | 34 passed ❌ | `× deal is guarded while an empty column exists` ✅ |

## Ninth guard, so it cannot come back

It already came back once: I reintroduced this exact pattern in tests I wrote hours earlier (#4450), and only mutation testing caught it. So `bun run check` now fails the build on it.

Two design decisions taken from the unmount guard's three rewrites in #4448:

- **Only mocks of a game API.** A plain `vi.fn()` prop callback *is* invoked synchronously and needs no flush — flagging those would be noise, and a noisy guard gets switched off.
- **Advancing fake timers counts as a flush.** `usePokerGame`'s debounce test drives its pending timer with `vi.advanceTimersByTime(300)`, and its assertion is meaningful precisely because of that. The guard flagged it until taught otherwise — exactly the false positive that gets a guard deleted rather than fixed.

Negative-controlled on both forms: removing a flush from a hook test (`act`) and from a page test (raw `.click()`) each exit 1 naming the exact line.

## Verification

- `bun run test` — **16,081 passed**
- `bun run typecheck`, `bun run check` (all 9 guards) clean
- Three mutations above confirmed failing after the change and passing before it
- Test-only, plus the guard script; no production files touched

### DIFF
--- a/frontend/scripts/check-design-tokens.mjs
+++ b/frontend/scripts/check-design-tokens.mjs
@@ -510,3 +510,62 @@ if (unmountViolations.length > 0) {
 }
 
 console.log(`unmount-safety: OK (${hookFiles.length} hooks; every await -> own-state write is guarded).`);
+
+// --- Vacuous "not called" assertions in tests (issues #4439, #4451) ---------
+//
+// `expect(<apiMock>).not.toHaveBeenCalled()` placed straight after an interaction
+// passes whether or not the action fired: `exec` dispatches through react-query's
+// `mutateAsync`, which invokes the mutation function on a microtask. The assertion
+// runs first and sees nothing, so it asserts nothing. #4443 fixed 148 of these and
+// found a real 3-page bug doing so; #4451 fixed 76 more that its detector could not
+// see, because that one keyed on `fireEvent` and missed the `act(...)` form used in
+// hook tests. The fix is `await flushPendingDispatch()` before the assertion.
+//
+// Checked per site, and the interaction list is deliberately broad — three widenings
+// (fireEvent -> act -> raw .click()) each turned up sites the previous pass called
+// clean, so a narrow matcher here reads as coverage it does not provide.
+const vacuousViolations = [];
+const INTERACTION = /fireEvent|userEvent|\bact\(|\.click\(\)|\.type\(/;
+async function collectTests(dir) {
+  const out = [];
+  for (const entry of await readdir(dir, { withFileTypes: true })) {
+    const full = join(dir, entry.name);
+    if (entry.isDirectory()) out.push(...(await collectTests(full)));
+    else if (/\.test\.tsx?$/.test(entry.name)) out.push(full);
+  }
+  return out;
+}
+for (const file of await collectTests(SRC_DIR)) {
+  const text = await readFile(file, 'utf8');
+  // Only mocks of a game API — a plain `vi.fn()` prop callback IS synchronous and
+  // needs no flush, so flagging those would be noise that gets the guard switched off.
+  const apiMocks = new Set([...text.matchAll(/const (\w+) = vi\.mocked\((\w+Api)\.\w+\)/g)].map((m) => m[1]));
+  if (apiMocks.size === 0) continue;
+  const lines = text.split('\n');
+  for (let i = 0; i < lines.length; i += 1) {
+    const m = /^\s*expect\((\w+)\)\.not\.toHaveBeenCalled/.exec(lines[i]);
+    if (!m || !apiMocks.has(m[1])) continue;
+    for (let j = i - 1; j >= Math.max(0, i - 8); j -= 1) {
+      // Already flushed. Advancing fake timers counts: usePokerGame's debounce test
+      // drives the pending timer with `vi.advanceTimersByTime(300)` rather than an
+      // await, and its assertion is meaningful precisely because of that — flagging
+      // it would be the guard crying wolf.
+      if (/\bawait\b|advanceTimersByTime|runAllTimers|runOnlyPendingTimers/.test(lines[j])) break;
+      if (INTERACTION.test(lines[j])) {
+        vacuousViolations.push({ file: relative(ROOT, file), line: i + 1, mock: m[1] });
+        break;
+      }
+    }
+  }
+}
+
+if (vacuousViolations.length > 0) {
+  console.error('\nVacuous "not called" assertions (no await between the interaction and the assertion):\n');
+  for (const v of vacuousViolations) {
+    console.error(`  ${v.file}:${v.line}  expect(${v.mock}).not.toHaveBeenCalled() cannot fail here`);
+  }
+  console.error('\nAwait flushPendingDispatch() first. See issues #4439 and #4451.');
+  process.exit(1);
+}
+
+console.log('vacuous-assertions: OK (every "not called" assertion on an api mock is awaited first).');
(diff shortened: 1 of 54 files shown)


yuta-yoshinaga:
Good catch on the boundary — fixed in b6fab197d, and it is the same mistake I had already fixed once in this session.

The unmount-safety guard added in #4448 needed exactly this boundary for exactly this reason: its window spilled past the end of `handleHint` and flagged a synchronous `setHint(null)` in `handleUndo`. I fixed it there and then failed to carry the lesson across when writing the second guard.

Reproduced it before fixing rather than taking it on faith — a two-test probe file where the first test uses `act()` and the second legitimately asserts "not called" on mount:

- without the boundary check: `boundary.test.ts:16 expect(mockExec).not.toHaveBeenCalled() cannot fail here` ← false positive, would fail the build for correct code
- with it: clean

Also confirmed the fix does not weaken real detection: removing a flush from `useUltiGame` still exits 1 naming the exact line. No behaviour change for the 76 sites in this PR, since all of them have the flush immediately above the assertion.

Folded into this branch rather than a follow-up, since it is the same guard.

PR #4456, opened 2026-07-28
feat(workers): add the fifth and sixth size buckets (ADR-0036 Phase 1)

yuta-yoshinaga:
Phase 1 of [ADR-0036](../blob/develop/docs/adr/0036-fifth-sixth-worker-capacity.md). Adds `extra2` and `extra3` as empty buckets that build and pass the size check. **No game moves** — Phase 2 does the rebalancing.

## Why now

| Worker | games | gzip | of 1 MB | headroom |
|---|---|---|---|---|
| casino | 64 | 1,032,505 | **98.5%** | 16.1 KB |
| classic | 57 | 1,037,964 | **99.0%** | 10.6 KB |
| solo | 52 | 980,329 | 93.5% | 65.1 KB |
| extra | 46 | 1,029,817 | **98.2%** | 18.8 KB |

~110 KB between them. The 45 proposed games (#4375–#4419) have nowhere to go — the same wall ADR-0032 hit at 179 games.

## Two measurements that changed the plan

**An empty worker is 232 KB gzip.** That is the Go runtime, `syumai/workers` and the shared helpers, paid once per worker regardless of content. I only learned this by building the empty buckets.

Separating it out gives the real marginal cost:

| Worker | gzip − baseline | per game |
|---|---|---|
| casino | 800,489 | 12,508 B |
| classic | 805,948 | 14,139 B |
| solo | 748,313 | 14,391 B |
| extra | 797,801 | 17,344 B |

**14.4 KB/game, not the 18.6 KB** I quoted earlier from dividing total gzip by total games — that charges the fixed baseline to the games and overstates by ~30%. It changes the answer to "how many more fit": ~57 per worker rather than ~46, and ~120 games of room across six.

**And it is the honest argument against this approach.** Six workers means paying that baseline six times — ~1.4 MB of duplicated runtime. Splitting to buy capacity gets *worse* the more you do it, so the ADR records that a seventh worker should trigger reconsidering the paid plan (10 MB) rather than splitting again.

## Local measurement now works

ADR-0032 said TinyGo could not be measured locally and rebalancing had to be driven off CI. That is no longer true. `docs/cloudflare-workers.md` now carries the procedure, including the two traps that cost me time:

- **`GOTOOLCHAIN=local` is required.** `go.mod`'s `toolchain go1.26.0` upgrades Go 1.25 to 1.26, and TinyGo 0.40.1 refuses: `requires go version 1.19 through 1.25, got go1.26`. CI sets it for the same reason.
- **Read sizes after `wasm-opt`.** It takes `extra` from 1,077,248 → 1,029,817. Judging on the pre-optimisation number would report a worker as over the limit when it is not.

## A trap in the scaffolding

Copying `wrangler.toml` from `extra` brought **its real KV namespace IDs** with it. Deployed as-is, two workers would have shared one session store and collided on keys. They are now `REPLACE_ME_*` markers.

## What this PR cannot do

Deploying needs three things only the account owner can do, listed in the ADR:

1. Create KV namespaces (production, preview, staging) and paste the IDs into the two `wrangler.toml` files
2. Set `WORKER_EXTRA2_URL` / `WORKER_EXTRA3_URL` and the `_STAGING_URL` variants
3. Set `VITE_WORKER_EXTRA2_URL` / `VITE_WORKER_EXTRA3_URL` for the frontend build

Until then the workers build and size-check in CI but are not deployed. That is what makes Phase 1 safe to merge on its own.

The ADR also records the **11 code touchpoints** for adding a worker — ADR-0032 documented moving a game between workers but not adding one, so this had to be rediscovered.

## Verification

- Both new workers build locally: **232,016 / 232,015 B gzip** (22% of the limit)
- Go suite green, `golangci-lint` **0 issues**
- `bun run test` — 16,086 passed; typecheck and all 9 guards clean
- No build artifacts staged (`workers/*/build/` is gitignored — checked)

Refs #4375


yuta-yoshinaga:
Fixed in ebfa44601 — you were right, and this invalidated the safety claim in my PR body.

I checked that the build-only workflow was safe to extend and **assumed** the deploy one was too. Tracing it as you describe: `deploy-cloudflare.yml` triggers on push to develop for `cmd/workers/**` and `workers/**` — which this PR touches — so the extra2/extra3 legs would `wrangler deploy` against `REPLACE_ME` KV IDs and fail. `fail-fast` defaults to true (confirmed: not overridden anywhere in the file), cancelling the other four legs, and `deploy-pages` is `needs: deploy-workers` at line 98. So the outcome is not 'the new workers stay undeployed', it is 'the deploy pipeline breaks' — the opposite of what I claimed made Phase 1 safe to merge alone.

Took your first suggestion rather than `fail-fast: false` + `continue-on-error`, for the reason you gave: two permanently-red legs is worse than deferring, and a red leg that is *expected* to be red trains people to ignore the deploy job.

Also done:
- `clean-workers` now covers `workers/extra{2,3}/build`
- `deploy-workers` stays at four but is now **annotated** with why — unannotated it reads as an oversight and the next person 'fixes' it into a broken deploy
- ADR-0036's touchpoint list said "deploy-cloudflare.yml — matrix and URL variables" without distinguishing the two workflows, which is precisely the trap I fell into. It now spells out the hazard and says the matrix entry belongs in the same change that fills in the real KV IDs.

One datum worth recording from this PR: CI reports `extra2: 594139 bytes raw, 232016 bytes gzip` — **byte-identical** to my local build. So the local TinyGo setup reproduces CI exactly, and Phase 2's rebalancing can be driven locally instead of a ~20 minute CI round trip each time. (The 237,643 in the artifact step is the upload zip, not the gzip figure.)

PR #4662, opened 2026-08-03
feat(tarocchini): add Tarocchini, the 262nd game

yuta-yoshinaga:
Closes #4409

4-player Bolognese trick-taker in **fixed 2v2 teams** (opposite seats partner) on the 62-card Tarocco Bolognese deck: 4 suits × 10 (the low pips 2–5 are absent) + 21 trumps + the Matto. No bidding.

## The rule that makes this game different

**The four papi rank equal, and the later-played one wins the trick.**

Every other trick-taker in this repo resolves with "update the winner only on a strictly stronger card." Borrowing that here hands the trick to whoever played their papa *first*. Two changes are needed and both are load-bearing:

1. The four papi collapse to a **single rank** — their raw values 2–5 would otherwise be four distinct ranks, so they could never tie.
2. The comparison **updates on a tie** between two papi.

Negative-controlled: removing the tie clause fails the papi tests, and restoring it is required to pass.

They are a **middle** group, not the top trumps — a higher ordinary trump still beats them. The CUI prints `パパ` rather than the number and the web face uses a distinct colour, because a numbered papa reads as "2 is weaker than 3".

## Two issue premises did not survive checking

Both recorded on #4409 before implementing:

- **"Deal 62 cards to 4 players" cannot be literal**: 62 ÷ 4 = 15.5. Real Tarocchini deals 15 each and gives the dealer the remaining 2 to bury (scarto), so `TarocchiniSurplus` is *derived* from the deck size rather than written as a number. The buried cards score for the dealer's team, which is why trumps and the Matto may not be discarded.
- **The issue calls the papi "the highest tarots."** Tarocco Bolognese has two separate groups of four — the equal-ranking papi just above the Begato, and the four distinct top trumps (Angelo/Mondo/Sole/Luna) — and the parenthetical looks like a conflation of them. I'd repeated that claim in my first comment before catching it. The papi set is now a single named constant pair with the uncertainty stated in the comment, so **re-identifying them is a one-line change**. If you have a source, say the word.

## This week's lessons applied up front

Ganjifa (#4661) and Vira (#4660) each cost a fix round for the same three things. Handled here from the start:

[...]

PR #4663, opened 2026-08-03
fix(domain): reseed rng on JSON restore, fixing a Tonk crash in production

yuta-yoshinaga:
Found by generalising a review finding on #4662 (Tarocchini). **This is a live crash in a shipped game.**

## The bug

**Tonk panics on the Cloudflare Worker whenever CPU difficulty is "Easy".**

`Tonk.UnmarshalJSON` never restores `rng`, and the Worker rebuilds the game from KV on every request without calling `SetRand`. `cpuDraw`'s Easy branch then calls `g.rng.Intn(3)` on a nil `*rand.Rand`:

```go
default:
    shouldPickDiscard = g.rng.Intn(3) == 0   // nil on every request after the first
```

It does **not** reproduce on the standalone server, which keeps the interactor in memory — so local runs and CI are both green while the deployed game crashes. Easy is selectable straight from the settings dropdown, so this is a default-adjacent path, not an edge case.

Reproduced with a test that restores from JSON and drives a CPU turn **without** calling `SetRand` (the line the real restore path does not have).

## Scope

| game | reseeds on restore? | rng used outside a guarded `shuffle()`? | status |
|---|---|---|---|
| **Tonk** | ❌ | ✅ `cpuDraw` | **crashes in production** |
| **Ganjifa** | ❌ | — | safe today, one call site from the same crash |
| **Vira** | ❌ | — | safe today, same |

Ganjifa and Vira are fixed here too so the domain is safe *by construction* rather than by coincidence.

Auditing the domain showed **the convention already existed** — Cuckoo, EgyptianRatscrew, Kemps, SixCardGolf, Slapjack, Spoons, ThirtyOne and Yaniv all reseed on restore. Three games had drifted from it and nothing enforced it. (Tarocchini #4662 had drifted too; fixed on its own branch.)

## The guard

`TestEveryRestoredGameReseedsItsRng` enforces the convention from the source: a domain with an `rng *rand.Rand` field and an `UnmarshalJSON` must assign `g.rng` there.

A per-game test would not have helped. The omission is **one missing line per game**, and the symptom only appears on the Worker — so the next game added repeats it. That is exactly what happened: I wrote the Ganjifa RNG-seeding guard earlier this week, documented it in the Tarocchini PR description, and still shipped Tarocchini and Minchiate with the same hole, because that guard covered `shuffle()` and not the restore path.

The guard asserts it inspected **more than eight** games, so a regex that drifts from the implementation fails loudly instead of silently passing on an empty set — the failure mode I hit with `check-hint-reasons.mjs` earlier this week.

It currently inspects **11** restorable games (Cuckoo, EgyptianRatscrew, Ganjifa, Kemps, SixCardGolf, Slapjack, Spoons, ThirtyOne, Tonk, Vira, Yaniv). The four equity engines have no `UnmarshalJSON` and are correctly skipped rather than miscounted.

## Relation to the vision

> **世界中のあらゆるトランプゲームを、誰でも無料で遊べるようにする。**

"無料で" means the Cloudflare free tier, and that is precisely the environment where this crashes. A game that panics on the deployed target is not playable, so this is a direct regression against the vision rather than a cosmetic issue — hence a standalone fix ahead of the feature branches.

## Test plan

- [x] `go test -tags test ./internal/...` — green
- [x] Reproduction fails before the fix (nil pointer dereference), passes after
- [x] Negative-controlled: removing Tonk's reseed fails both the regression test and the guard
- [x] Guard confirmed to inspect 11 restorable games, not an empty set
- [x] CI — 18/18

## Follow-up, not in this PR

Seeding style is split evenly across the domain — 9 files use `time.Now().UnixNano()`, 9 use `rand.Int63()`. The former can hand two games constructed in the same nanosecond an identical deal. Tonk now has one of each (constructor vs. restore). That is a pre-existing, repo-wide tidy-up rather than part of this crash fix, so it is filed separately instead of widening a hotfix diff.

### DIFF
--- a/internal/domain/Ganjifa.go
+++ b/internal/domain/Ganjifa.go
@@ -252,6 +252,11 @@ func (g *Ganjifa) UnmarshalJSON(data []byte) error {
 	g.gameEndFlag = j.GameEndFlag
 	g.winnerPlayer = j.WinnerPlayer
 	g.actionLog = j.ActionLog
+	// **復元したら必ず乱数源を張り直す。**Cloudflare Worker は毎リクエスト KV から
+	// 組み直すので SetRand は一度も呼ばれない。rng を nil のままにすると、
+	// シャッフル以外で rng を使う経路 (CPU の乱択など) が nil デリファレンスで
+	// 落ちる。呼び出し側ごとにガードするのではなく、ここで構造的に潰す。
+	g.rng = rand.New(rand.NewSource(rand.Int63()))
 	return nil
 }
 
--- a/internal/domain/Tonk.go
+++ b/internal/domain/Tonk.go
@@ -856,5 +856,10 @@ func (g *Tonk) UnmarshalJSON(data []byte) error {
 	}
 	g.isTonk = j.IsTonk
 	g.isUndercut = j.IsUndercut
+	// **復元したら必ず乱数源を張り直す。**Cloudflare Worker は毎リクエスト KV から
+	// 組み直すので SetRand は一度も呼ばれない。rng を nil のままにすると、
+	// シャッフル以外で rng を使う経路 (CPU の乱択など) が nil デリファレンスで
+	// 落ちる。呼び出し側ごとにガードするのではなく、ここで構造的に潰す。
+	g.rng = rand.New(rand.NewSource(rand.Int63()))
 	return nil
 }
--- a/internal/domain/TonkRng_test.go
+++ b/internal/domain/TonkRng_test.go
@@ -1,0 +1,40 @@
+//go:build test
+
+package domain
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// **復元したゲームで Easy の CPU を動かす。**Cloudflare Worker は毎リクエスト
+// KV から組み直すので `SetRand` は一度も呼ばれない。`UnmarshalJSON` が rng を
+// 戻さないと、`cpuDraw` の Easy 分岐 (`g.rng.Intn(3)`) が nil の *rand.Rand を
+// 触って落ちる —— サーバー実行ではインタラクタがメモリに残るため再現しない。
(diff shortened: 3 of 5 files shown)


yuta-yoshinaga:
レビューありがとうございます。**2つの nit のうち1つは自分の記述ミスでした。**

## 「9 restorable games」は誤り → 11

ご指摘のとおりです。ガード自身に出力させて確認しました:

```
rng_restore_test.go:60: checked 11 restorable games with a struct rng
--- PASS: TestEveryRestoredGameReseedsItsRng
```

**数えずに書いた数字でした。**ガードは自分が「何件見たか」を出力するようにしてあるのに、それを走らせずに PR 説明へ 9 と書いています。PR 説明を 11 に修正し、内訳（Cuckoo / EgyptianRatscrew / Ganjifa / Kemps / SixCardGolf / Slapjack / Spoons / ThirtyOne / Tonk / Vira / Yaniv）と、equity エンジン4本が `UnmarshalJSON` を持たないため正しく skip されている点も明記しました。

## シード方式の不統一 → 別対応にします

Tonk が `time.Now().UnixNano()`（コンストラクタ）と `rand.Int63()`（復元）の両方を持つ件、事実そのとおりです。ドメイン全体では **9ファイルが time 系・9ファイルが `rand.Int63()` 系**で真っ二つでした。

`time.Now().UnixNano()` には**同一ナノ秒に作った2局が同じ配りになる**という実害があるので直す価値はありますが、**この PR の目的は本番クラッシュの停止**です。関連して見える別の変更を混ぜるとホットフィックスの「安全に入れられる」性質が薄まるので、**リポジトリ横断の整理として別途切り出します**。この PR は最小のまま維持し、PR 説明にも follow-up として明記しました。

## 実行できなかったテストについて

サンドボックスで `go test` / `golangci-lint` が実行できなかったとのこと、CI は **18/18 グリーン**（codecov も "All modified and coverable lines are covered by tests"）です。ローカルでも全 Go スイート・負のコントロール両方向を確認済みです。

マージします。

PR #5405, opened 2026-08-15
feat(cui): translate the rest of the parser's argument rejections

yuta-yoshinaga:
`Refs #5384`. Stacked on #5404 -- **merge that first**.

## Measured

English replies in Japanese mode, over every argument-taking command given a bad argument:

| | count |
|---|---|
| before #5404 | 174 |
| after #5404 | 119 |
| after this | **49** |

## What moved

148 call sites off `cuiutil.WithParsedInt` / `ParseIntArg`, behind 185 generated keys.

Only the **93 nouns** and the wordy hints are translated by hand. The sentence frames
and the numeric hints are generated: `1-13` reads the same in both languages, and
hand-writing 185 sentences invites 185 typos. The generator writes nothing unless
every wording resolves *and* no two wordings collide on one key -- a collision would
silently give one message the other's text, which is invisible in review.

Nine keys already existed with different English. Reusing `anteRequired` would have
added `(e.g. sa 10)` to games whose command is not `sa`, so those took a distinct name.

## Three things worth reading

1. **`m.On("Surrender")` got rewritten to a message key.** `"Surrender"` is a prefix
   of `"Surrender rule is required."`, and the fragment rule cannot tell a mock method
   name from a message. One occurrence; it panicked rather than passing quietly.
2. **`--- FAIL` counted zero while the package did not build.** The verdict is the last
   line of `go test`, not the count of `--- FAIL`.
3. **41 assertions looked for the bare word `"required"` or `"Invalid"`.** That is the
   same defect one layer up, so they assert the reply carries the rejection marker now
   -- except where the reply is still a raw English return, where the English literal is
   what the code produces and pinning it is honest.

## Not in this PR

The remaining 49 come from raw returns like `fmt.Sprintf("Invalid row: %s.", …)`.
**Only 10 of 103 fit any frame** -- the rest are bespoke sentences
(`"Bid is required (petite, garde, gardesans, or gardecontre)."`), so they need
translating by hand, not converting by script.

## Test plan

- [x] `go test -tags test ./internal/adapter/controller/` -- green
- [x] `golangci-lint run ./internal/adapter/controller/...` -- 0 issues
- [x] full `go test -tags test ./...`
- [x] English-reply probe re-measured on this tree: 119 -> 49


yuta-yoshinaga:
Must-fix は正しい指摘でした。**#5406 で解消しています**（この PR ではなく次の PR で直した理由を一行）。

`execSetSeats` の `len(args) == 0` ガードは、この PR が対象にした `cuiutil` 呼び出しではなく生の return なので、この PR のスコープ（"cuiutil の呼び出し口"）の外側でした。#5406 がその生 return 群を担当し、そこで `invalidArg("numberOfSeatsRequired469")` に変換済みです。

指摘のおかげで**もう一つ**見つかりました: 変換した結果 `numberOfSeatsRequired469` と `numberOfSeatsRequired46Or9` が同じ文を返す二重キーになっていたので、#5406 で 1 つに畳みました。あわせて `ss` 引数なしのテストが**そもそも 1 件も無かった**（だから死んだキーが動いて見えた）ので、レンダリング結果と突き合わせるケースを追加しています。

PR #5407, opened 2026-08-15
fix(cui): refuse a mistyped optional argument instead of defaulting it

yuta-yoshinaga:
`Refs #5390`. Stacked on #5406 -- **merge that first** (it carries the keys this uses).

## Measured, 20 deals per command

`ParseOptionalInt` could not tell *absent* from *unparseable* and returned the default
for both. Six commands executed a move the player never typed, every single time:

| game | command | executed |
|---|---|---|
| blackjack | `b 100 abc` | 20/20 — Pair Plus silently becomes 0 |
| sevens | `p abc` | 20/20 — plays card 0 |
| speed | `p abc` | 20/20 — plays card 0 |
| speed | `p 0 abc` | 20/20 — plays onto pile 0 |
| speed | `sd abc` | 20/20 — resets the game at Normal |
| sevens | `p 0 abc` | 20/20 |

**Five are fixed** (re-measured: 0/20). `ParseOptionalIntKeys` keeps the default for an
absent argument and refuses a present-but-unparseable one.

## blackjack's is the one that costs money

`b 100 abc` staked the main bet and quietly set the Pair Plus to nothing. The hand plays
out as if the player had never asked for the side bet; nothing on screen says otherwise
and only the balance disagrees.

## Three tests were named after the bug

```go
t.Run("bet with invalid PP (ignored)", ...)
t.Run("bet with PP and invalid T3 (ignored)", ...)
t.Run("bet with invalid handCount (ignored)", ...)
```

They assert the refusal now. A test named after the bug is how the bug survives.

## The sixth is a different bug

`sevens p 0 abc`: `p` reads one argument and the second is **surplus, not defaulted**.
Refusing surplus arguments is a policy decision for every command at once, so it is not
folded in here.

## Test plan

- [x] `TestParseOptionalIntKeys` — absent takes the default, index past the end takes the
      default, a valid argument wins, a typo is refused **and marked**
- [x] `go test -tags test ./internal/adapter/controller/... ./internal/infrastructure/ui/`
- [x] `golangci-lint run ./internal/adapter/controller/...` — 0 issues
- [x] probe re-measured on this tree: 6 commands → 1

### DIFF
--- a/internal/adapter/controller/BlackJackCuiController.go
+++ b/internal/adapter/controller/BlackJackCuiController.go
@@ -71,9 +71,18 @@ func (bcc *BlackJackCuiController) Exec(command string) string {
 				if !ok {
 					return errMsg, true
 				}
-				ppBet := cuiutil.ParseOptionalInt(args, 1, 0)
-				t3Bet := cuiutil.ParseOptionalInt(args, 2, 0)
-				handCount := cuiutil.ParseOptionalInt(args, 3, 0)
+				ppBet, errMsg, ok := cuiutil.ParseOptionalIntKeys(args, 1, 0, "invalidPairPlusBet")
+				if !ok {
+					return errMsg, true
+				}
+				t3Bet, errMsg, ok := cuiutil.ParseOptionalIntKeys(args, 2, 0, "invalidTripsBet")
+				if !ok {
+					return errMsg, true
+				}
+				handCount, errMsg, ok := cuiutil.ParseOptionalIntKeys(args, 3, 0, "invalidHandCount")
+				if !ok {
+					return errMsg, true
+				}
 				return bcc.bji.Bet(amount, ppBet, t3Bet, handCount), true
 			case "ssr", "setsurrenderrule":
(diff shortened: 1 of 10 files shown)


yuta-yoshinaga:
inline の 2 件（Speed / Sevens に呼び出し口のテストが無い）は**正しい指摘で、レビュー後の 2 コミットで対応済み**です。auto-review は push で再実行されないため、レビューは最初のコミットに対するものになっています。

- `9a063ca97` — Speed 4 ケース / Sevens 3 ケース。**拒否とセットで「省略形が壊れていないこと」も見ています**（`p` 引数なしは sevens でパス、speed では 0 番を 0 番の山へ）。拒否側だけ見るテストは、拒否を直したついでに既定値を消しても緑のままです
- `610d8ed5d` — codecov が指した Sevens の `j` の 1 番目と 3 番目の引数。真ん中（スート）だけテストがあり、両端は既定値に落ちても緑でした

**負のコントロール**: `ParseOptionalIntKeys` を既定値に戻す実装へ書き換えると 7 件中 5 件が落ちます。存在理由どおりに落ちることを確認済みです。

codecov も `All modified and coverable lines are covered by tests` になっています。

PR #6186, opened 2026-08-21
feat(fortress): add Fortress solitaire game

yuta-yoshinaga:
Closes #5468

Second game of the #5422–#5470 batch.

## What Fortress is

A Beleaguered Castle variant, cloned from that domain and diverging in three ways:

| | foundations | tableau layout | build rule |
|---|---|---|---|
| Beleaguered Castle | 4 Aces pre-seeded | 8 columns × 6 | any suit, descending |
| **Fortress** | **start empty** | **10 columns** (2×6 + 8×5) | **same suit, ascending OR descending** |

Single-card moves and "any card to an empty column" carry over unchanged.

## Bucket: solo, not the roomiest worker

`extra4` has more headroom (142.6 KB vs solo's 105.8 KB), but Fortress clones a **solo-tagged** domain — `BeleagueredCastleTableauCard` and friends are already local to that bucket. Putting it in extra4 would have required the same four bucket-local type copies Alaska needed. solo has ample room for one ~12 KB game, so this is the cheaper correct answer. When solo tightens, later solitaires will pay the duplication instead.

## Three cloned fixtures were rebuilt, not re-asserted

This is the trap from #6185, and it bit harder here because the column count changed too:

- **"No legal move" fixtures filled 8 columns** — all of Beleaguered Castle's, but only 8 of Fortress's **ten**. An empty column accepts any card, so a move always existed and the stalemate was unreachable. `fortressDeadTableau()` fills all ten and documents why each of its three conditions is load-bearing.
- **Foundation fixtures assumed Reset seeded the Aces.** Fortress deals the whole deck, so `seedFortressAce()` places one explicitly.
- **Tableau fixtures paired mixed suits**, which Beleaguered Castle allowed.

Every new rule assertion has a negative control — adjacent-rank-wrong-suit and same-suit-two-apart both stay rejected, so "same suit, either direction" cannot quietly degrade into "same suit, anything". The dead-tableau fixture has one too: add one legal move, the hint must come back.

## Frontend touchpoints Alaska did not have

Different games carry different helper structures, so enumerating the source game's full footprint beats reusing a checklist:

- `hooks/useFortressGame.ts` (+ test)
- `utils/fortressLegalTargets.ts` (+ test)
- `utils/cli/formatters/fortressFormatter.ts` (+ test)
- `api/games/mao.ts` — an unexpected shared re-export point for a batch of solitaire `MoveZone` types

## Verification

- `GOOS=js GOARCH=wasm go build -tags <w>` — all 7 workers OK
[...]

PR #6187, opened 2026-08-21
feat(somerset): add Somerset solitaire game

### LINKED ISSUE #5469: Somerset (サマセット) の追加
## 概要 (Overview)
Somerset は全52枚を最初から表向きの10列に配るオープンパッカー系ソリティアで、降順・異色でしか積めずフリーセルも山札も無いため、一手ごとの先読みが結果を左右する点が特徴。

- **メカニクス**: 既存の `bakersgame` は10列全表向きで同スート限定の降順にしか積めない一方、`freecell` は異色降順ルールだがフリーセル4つを持つ。Somerset は積み方が異色降順という点で `freecell` に近いが、`bakersgame` 同様フリーセルを一切持たず、さらに移動は常に1枚のみに制限される点が両者のどちらとも異なる。
- **地域カバレッジ**: FreeCell系譜の中でも「退避先ゼロ」という最も厳しい難易度帯を補う。
- **実装の見通し**: `internal/domain/FreeCell.go`（`NewBakersGame` の10列全表向き初期配置）を土台に流用できる。新規に書くのは、フリーセルを持たない構成への差し替え、積み方を同スート限定から異色限定へ変更する判定、常に1枚移動のみを許可する制約。

## 仕様案 (Requirements)
- **人数**: 1人
- **デッキ**: 52枚
- **主要ルール**:
  1. 52枚をすべて表向きで10列に均等に配る（フリーセルは用意しない）。
  2. ファンデーション4つはスートごとにAから昇順に積んで完成させる。
  3. タブロー上では異なる色かつ降順（例: 赤7の上に黒6）でのみカードを接続できる。
  4. 移動は常に1枚のみで、複数枚をまとめて動かすことはできない。
  5. 空いた列にはどのカードでも置ける。
  6. 補充する山札は無く、52枚すべてがファンデーションに収まればクリア。

- 再利用: `internal/domain/FreeCell.go` の `NewBakersGame` 由来の10列全表向き初期配置を流用し、新規に書くのはフリーセル無し構成、異色降順判定、1枚移動限定の制約のみ。
- 新規ドメイン型: `internal/domain/Somerset.go`（単独プレイのソリティアのため Config/Player は不要）

---
_新規ゲーム候補バッチ（2026-08-16, 49件）により起票。実装時は docs/new-game-checklist.md に従い、バケットは着手時に .claude/skills/rebucket-game/scripts/measure.sh で実測すること。ブロッカー: #5421（6 worker の gzip 残容量が合計 220 KB しかなく、7番目の worker 新設が前提）（スラッグ案: `somerset`）_

yuta-yoshinaga:
Closes #5469

Third game of the #5422–#5470 batch.

## Cloned from Fortress, not FreeCell

The issue suggests reusing `FreeCell.go`. But Fortress (merged in #6186) already has Somerset's exact layout — whole deck face-up over 10 columns, empty foundations, no free cells, single-card moves, any card to an empty column. Only the build rule differs:

| | tableau build |
|---|---|
| Fortress | same suit, ascending **or** descending |
| **Somerset** | **alternating colour, descending only** |

From FreeCell this would have meant stripping out free cells and supermove sizing to arrive at a layout that already exists. From Fortress it is one function. Bucketed `solo` to match, so no bucket-local type duplication.

## Fixtures fixed, not assertions relaxed

Four cloned fixtures paired same-suit cards — legal in Fortress, rejected here — and were switched to alternating colour. The dead-tableau fixture survives the rule change intact (all ten columns hold a 5, so no rank is ever adjacent), and its negative control still means something: the black 6 it introduces can legally accept a red 5.

Mutation-tested rather than trusting green:

| mutation | failing subtests |
|---|---|
| drop the alternating-colour check | 2 |
| allow ascending (Fortress's rule) | 1 |

## New: `clone-game.py`

This is the third near-identical clone and **45 remain**. Doing it by hand kept missing files, because the per-game helper set differs per game — RussianSolitaire has no `hooks/use<X>Game.ts`, Fortress has three utils Alaska never needed, and `MoveZone` types are re-exported through `api/games/mao.ts`. The script finds the source's real footprint by grep rather than a fixed checklist, and separates the ~34 per-game files from the ~17 shared registration points that need ordered insertion.

Two bugs found while writing it, both fixed and documented in its docstring:

- **Content grep alone misses `docs/manual/web/<game>.md`** — it is written entirely in Japanese and never spells the ascii key, so cloning Fortress by content silently skipped its Web manual. Now matches on filename **or** content.
- **The Japanese display name needs `--jp`.** Without it the page test fails with `Unable to find text: /<old name>/`, and it also survives in Go doc comments.

Generic helper names in cloned `_test.go` files (`cardSpec`) still collide in the shared `package_test` and need prefixing by hand — documented, not automated, since renaming them blindly would be riskier.

## Verification

- `GOOS=js GOARCH=wasm go build -tags solo|extra4` — OK
- `go test -tags test ./...` — green
- `golangci-lint run ./...` — 0 issues; `goimports` clean
- `count-audit.sh` — 321 games consistent
- `bun run typecheck` clean; Somerset frontend tests 29 + 36 pass
- guards: discover-blurbs / discover-axes / hint-coverage / hint-gate / cli-hint-reachable / message-codes all report 321
- CLI smoke test: 10 columns dealt 6/6/5×8, empty foundations
- `api/openapi.yaml` CRLF preserved — **136 added, 0 deleted**
- Self-reviewed the sweep classes (nested locales, mermaid node prose, manual contradictions) — clean


yuta-yoshinaga:
**Your `somersetLegalTargets` finding was correct — thank you.** The review truncated before posting the detail, so here is what it was, and it is worse than one file:

**The same defect is already on `develop`, in Fortress (#6186).**

Both files still carried Beleaguered Castle's rule — `card.value === top.value - 1`, with **no suit or colour test at all** — because the frontend duplicates the server-side rule and the clone chain (BeleagueredCastle → Fortress → Somerset) never updated it. So the destination preview highlighted columns the backend rejects: Fortress lit every adjacent-rank column regardless of suit; Somerset would have lit same-colour columns.

**Why nothing caught it:** the cloned tests *assert the wrong rule* rather than assert nothing. `it('accepts any suit onto a tableau card one rank higher')` passes happily against Beleaguered Castle's logic — it is a confidently wrong test, not a vacuous one. Domain-level mutation testing can't see it either, since the bug lives in the frontend copy rather than the domain.

Fixed in 7a31882e:

| file | now |
|---|---|
| `fortressLegalTargets.ts` | same suit, adjacent rank either direction |
| `somersetLegalTargets.ts` | alternating colour, descending only |

Both tests rewritten to assert the real rule with a negative control for every way it can be wrong — wrong suit/colour, wrong direction, two ranks apart. `FortressPage`'s fixture paired a spade with a heart, so its highlight test was encoding the old rule too; Somerset's fixture is legal by luck, but its comment claimed "スート不問", which would have misled the next clone.

I audited the four sibling `*LegalTargets.ts` files — `americanToad`, `bakersDozen`, `kingAlbert`, `beleagueredCastle` each still match their own domain. Only the two I cloned had drifted.

Both doc comments now state outright that the file duplicates the server rule and must change with it.

PR #6192, opened 2026-08-21
feat(perseverance): add Perseverance solitaire game

### LINKED ISSUE #5443: Perseverance (パーシビアランス) の追加
## 概要 (Overview)
Perseverance は Baker's Dozen 系のタブロー配置ソリティアで、4枚×13列に配った札を同スート降順で組み上げます。Baker's Dozen との決定的な違いは、行き詰まった際に場札を集めて最大2回まで再配りできる救済ルールがある点です。

- **メカニクス**: 既存の `bakersdozen`（4枚×13列、Kを列末尾へ移動して開始、再配り無し）とタブロー配置・K処理は共通ですが、Perseverance は再配りが無いという Baker's Dozen の最大の制約を取り払い、動けなくなった時点の場札を回収して再度配り直す操作を2回まで許します。これにより難度が大きく下がり、初心者向けのバリエーションとして位置づけられます。
- **地域カバレッジ**: 「忍耐強さ」を意味する名の通り、Baker's Dozen 系に救済手段を加えた派生形を追加し、同系統のルール難度の幅を広げます。
- **実装の見通し**: `internal/domain/BakersDozen.go` の4枚×13列の配り、Kを列末尾へ動かす初期化、同スート降順の組み上げ判定はそのまま流用できます。新規に書くのは、行き詰まり判定後に場札を回収してシャッフルせず（または規定順で）再配りするロジックと、再配り回数（残り2回）のカウント管理です。

## 仕様案 (Requirements)
- **人数**: 1人
- **デッキ**: 52枚
- **主要ルール**:
  1. 52枚を4枚ずつ13列に配る。
  2. 各列に含まれるKを列の末尾（一番下）へ移動する初期整理を行う。
  3. 各列の最上段の札を、他の列の最上段に同スート降順で重ねられる（空列にはどの札でも置ける）。
  4. 4スート×13枚（A→K昇順）の基礎札へ、条件を満たした最上段の札を送れる。
  5. 動かせる手が尽きたら、場に残る札を集めて再度13列に配り直せる（最大2回まで）。
  6. 3回目の手詰まりで敗北、全基礎札が完成すればクリア。

- 再利用: `internal/domain/BakersDozen.go` の配り・K処理・降順組み上げ判定を流用し、新規に書くのは再配り（最大2回）のロジックと回数管理のみ。
- 新規ドメイン型: `internal/domain/Perseverance.go`, `PerseveranceConfig.go`, `PerseverancePlayer.go`

---
_新規ゲーム候補バッチ（2026-08-16, 49件）により起票。実装時は docs/new-game-checklist.md に従い、バケットは着手時に .claude/skills/rebucket-game/scripts/measure.sh で実測すること。ブロッカー: #5421（6 worker の gzip 残容量が合計 220 KB しかなく、7番目の worker 新設が前提）（スラッグ案: `perseverance`）_

yuta-yoshinaga:
Closes #5443

Baker's Dozen variant, cloned from `bakersdozen`. Bucket **`extra4`**, chosen on measured gzip headroom (142.6 KB, against `solo`'s 70.3 KB — `solo` is now the tightest of the seven). `GOOS=js GOARCH=wasm go build -tags extra4` confirms the clone's lineage strands no `solo`-tagged symbol, so the cross-bucket move costs nothing.

## The issue's spec was wrong in four places

Checked against [goodsol](https://www.goodsol.com/games/perseverance.html) and [Wikipedia](https://en.wikipedia.org/wiki/Perseverance_(solitaire)), which agree with each other:

| Rule | Issue said | Actual | `BakersDozen` code today |
|---|---|---|---|
| Columns | 4×**13** | **12** columns of 4 | 13 ✗ |
| Foundations | (not mentioned) | **start holding their ace** | empty ✗ |
| Tableau build | same-suit descending ✓ | same-suit descending | **suit-blind** (`GetValue()-1`) ✗ |
| Run moves | (not mentioned) | **a same-suit run moves as a unit** | single card only ✗ |
| Empty columns | "any card may be placed" ✗ | **not fillable** | not fillable ✓ |
| Redeals | 2, no shuffle ✓ | 2, gathered in reverse order, no shuffle | none (new code) |

So the issue's "the deal, the king handling and the descending build can be reused as-is" holds only for the kings.

**Kings-to-bottom is a judgement call**: goodsol does it, Wikipedia is silent. Kept, since it matches the clone source and the issue, and it is the one thing they agree on.

## What that meant in code

- **Deal** — four aces lifted onto the foundations first, then 48 cards into 12 columns. Suit constants are 1–4 but foundation indices are 0–3, so the ace goes to `foundation[design-CardDesignSpade]`; indexing by `design` directly would have run off the end of the array on ♦.
- **`canPlaceOnTableau`** — gains the suit check.
- **`MoveTableauToTableau`** — a new `isRun` admits a descending same-suit run; a broken run moves nothing rather than partially.
- **`Redeal`** — piles gathered in reverse order (11 onto 10 onto 9…) and re-dealt four at a time **without shuffling**, so the next board is a function of the last one, not of luck. Two per game.

`redealsLeft` is threaded through both the undo snapshot and the KV wire format. Without the former, undo returns the board but keeps spending redeals; without the latter the stateless Worker hands out two fresh redeals every request (#4478's shape). A value outside 0–2 coming back from KV is refused rather than clamped.

## A clone's tests pass by default, because they assert the source game

**14 Go tests and 8 frontend tests were green while encoding Baker's Dozen** — `♥4` onto `♠5`, an ace sitting in the tableau, `LegalTargets` returning `[0, 1]` where suit now narrows it to one, `makeTableau` padding to 13. This is the same class that produced the review findings on #6191 and shipped once already in `fortressLegalTargets`; it is not caught by vacuous-test hunting, because the tests are confidently wrong rather than empty.

Every one is corrected and now carries a negative control the source game would fail (cross-suit descending must be **refused**; a mixed-suit group must **not** travel), so the file cannot go green again if the borrowed rule creeps back. My own first run-move fixture was wrong too (♠8 onto ♠10) and the test caught it.

## Also

- The `message-codes` guard caught six missing translations. Baker's Dozen's `stalemate` wording — "undo or give up" — is wrong advice in a game with a redeal button, so both locales now mention it.
- Both manuals rewritten against the implementation (they described 13 columns, suit-blind stacking and single-card moves).
- OpenAPI: +161/−0, CRLF verified intact (61,409 CRLF, 0 bare LF).

## Test plan

- [x] `go test -tags test ./internal/...`
- [x] `golangci-lint run ./internal/...` — 0 issues
- [x] `goimports -l` on every changed Go file — clean
- [x] `cd frontend && bun run check` — passes (`message-codes`, `discover-blurbs`, `discover-axes`, `guard-floors` all OK at 326 games)
- [x] `cd frontend && bun run typecheck`
- [x] 176 targeted frontend tests (Perseverance + every registration point touched)
- [x] `count-audit.sh` — ✅ all count assertions consistent (326 games)
- [x] `GOOS=js GOARCH=wasm go build -tags {extra4,solo,classic}` — no stranded symbols
- [ ] Full frontend suite + E2E — left to CI (4 local shards in parallel exhaust this box)


yuta-yoshinaga:
Review finding fixed in d47ecb2d0 — thank you, this was a real defect and the diagnosis was exactly right.

The run move was implemented in the domain, described in OpenAPI and both manuals, and reachable by nobody. All four surfaces are now wired:

- **Web** — Perseverance no longer uses the shared `dispatchTopCardMove`. That helper is *correctly* top-card-only for its three real users, so it was the wrong thing to reach for here; Perseverance has its own dispatcher forwarding `From.CardIndex`. Validity stays the domain's job — `isRun` refuses a broken run and moves nothing, so forwarding the client's index is not the untrusted-input problem the shared helper's doc comment warns about.
- **CUI** — new `m <col> <idx> <col>` form; the two-argument form still means "top card".
- **Page** — `perseveranceStartsRun` replaces the `isTop` gate for selection, dragging, cursor and destination preview.
- **CLI mode** — already parsed the index; it now actually reaches the domain.

You also spotted that `PerseveranceWebController_test.go`'s *"move tableau to tableau **ignores client cardIndex**"* pinned the wrong behaviour. It came straight from Baker's Dozen, where dropping the index is right. Inverted, and the mock now asserts index `1` arrives rather than `-1`, so the old implementation could not pass it.

**Your "minor" was load-bearing once the wiring landed.** `GetHint`/`checkStalemate` only scanned column tops — fine for a game that moves one card, wrong here: they would have declared a board stuck while a deeper run could still move. Both now walk `runStarts`, with a negative control (a run broken by suit is genuinely stuck).

That fix also caught a flake of my own: the new stalemate test read `IsStalemate()` straight after `SetTableau`, which does not recompute it, so it was asserting on whatever the shuffled deal left behind — green alone, red in the full-package run. It now performs the move and asserts on the recomputed value, on a board that keeps a follow-up move so it measures "not stuck" rather than "that was the last move". Three consecutive full-package runs are clean.

Gates re-run: `go test -tags test ./internal/...`, `golangci-lint` 0 issues, `goimports` clean, `bun run check`, `typecheck`, 68 targeted frontend tests, `extra4` WASM boundary.

PR #6195, opened 2026-08-21
feat(mrsmop): add Mrs. Mop solitaire game

### LINKED ISSUE #5441: Mrs. Mop (ミセス・モップ) の追加
## 概要 (Overview)
Mrs. Mop は Spider の完全公開版とも呼ばれるソリティアで、104枚（2組デッキ）すべてを最初から表向きでタブロー13列に配ります。伏せ札が一切無いため、配った瞬間に「その手が解けるかどうか」が理論上確定しているのが最大の特徴です。

- **メカニクス**: 既存の `spider`（2組104枚、Spider 系の同スート降順13枚完成で列除去）とタブロー構成・完成条件は同一ですが、Mrs. Mop は山札からの追加配りが無く、13列すべてが最初から表向きで公開される点が根本的に異なります。伏せ札という不確実性を排除することで、Spider の判定ロジックはそのまま使えても難度と戦略性が大きく変わります。
- **地域カバレッジ**: 完全情報化された Spider 系という、既存の `spider` の姉妹形を追加し、運要素を減らした遊び方の選択肢を増やします。
- **実装の見通し**: `internal/domain/Spider.go` の同スート降順13枚完成判定・列間移動バリデーション・列除去処理はそのまま流用できます。新規に書くのは、104枚を最初から全表向きで13列に配る初期化処理（Spider の伏せ札生成・山札からの追加配りロジックの除去）のみです。

## 仕様案 (Requirements)
- **人数**: 1人
- **デッキ**: 52枚 × 2組（104枚）
- **主要ルール**:
  1. 104枚をシャッフルし、13列すべてに配り切る（全枚数を均等割り、全て表向き）。
  2. 山札（stock）は存在せず、追加配りも無い。
  3. 各列の最上段の札は、他の列の最上段に「1つ小さいランク」を重ねられる（スート無視で移動可、ただし同スート連続でのみ完成扱い）。
  4. 同スートで昇順に連続した束は、まとめて他の列へ移動できる。
  5. ある列で同スート降順の連続13枚（K〜A）が揃うと、その列は場から取り除かれる。
  6. 8組（104枚 ÷ 13）すべてを取り除けばクリア、動かせる手が尽きれば手詰まり（伏せ札が無いため理論上は開始時点で解けるかどうかが決まっている）。

- 再利用: `internal/domain/Spider.go` の完成判定・列間移動・列除去ロジックをそのまま流用し、新規に書くのは山札を使わない全表向き初期配りロジックのみ。
- 新規ドメイン型: `internal/domain/MrsMop.go`, `MrsMopConfig.go`, `MrsMopPlayer.go`

---
_新規ゲーム候補バッチ（2026-08-16, 49件）により起票。実装時は docs/new-game-checklist.md に従い、バケットは着手時に .claude/skills/rebucket-game/scripts/measure.sh で実測すること。ブロッカー: #5421（6 worker の gzip 残容量が合計 220 KB しかなく、7番目の worker 新設が前提）（スラッグ案: `mrsmop`）_

yuta-yoshinaga:
Closes #5441

**Open Spider**: two full decks dealt into **13 columns of 8, every card face up**, with no stock and no extra deal. Build down regardless of suit, move same-suit runs as a unit, discard complete same-suit King-to-Ace sequences, clear all 8 to win. Bucket **`extra4`** on measured headroom; the WASM boundary check is clean.

Sources: [Wikipedia](https://en.wikipedia.org/wiki/Mrs._Mop), [goodsol](https://www.goodsol.com/games/mrsmop.html), [PySol](https://pysolfc.sourceforge.io/doc/rules/mrsmop.html) — **and the issue is right this time**, which is worth saying after four consecutive specs that weren't.

## Clone source: Spider, and this time adapting the tests was correct

By the "what would I delete" test: the stock and face-down dealing are *mechanics*; the objective — build same-suit K→A runs and remove them — is **identical**. That has a consequence I haven't hit before in this batch: **nearly every cloned Spider rule test states a rule Mrs. Mop genuinely shares**, so rewriting them from scratch (as #6193 and #6194 required) would have thrown away correct work. Only the deal, the stock and the face-down cases needed to change.

## The default was the real defect

Inheriting Spider's **1-suit default** would mean the board a player first opens is *not Mrs. Mop*. It now defaults to the proper 4-suit game — **including for an unrecognised config value**, which previously degraded into a different game rather than the right one.

1- and 2-suit survive as explicitly-labelled *relaxed variants*. I considered deleting them outright for faithfulness, but the vision's "anyone can play" goal justifies keeping them: a 4-suit board is brutal. They're documented as variants, not as Mrs. Mop.

## Everything the stock touched is gone, not zeroed

Domain field, `Deal` across interface/interactor/both controllers, the stock row and deal button, the `d` keybinding, the CLI `deal` command, the header's stock/remaining-deals line — **and the "cannot deal, empty column exists" guard with its shake animation**. That last one matters: an empty column in Mrs. Mop is not an obstacle blocking a deal, it's a **free slot that accepts anything**, so keeping the warning would have actively misled players about a rule that works the opposite way.

`GetStockCount()` remains, returning 0, only because the shared interface requires it — with a comment saying so.

## Smaller things the clone got wrong

- **`spiderMovableRun` is reused, not cloned.** The run rule is identical; the clone had rewritten the import to `mrsMopMovableRun`, a name that does not exist. Correct reuse of a shared helper, not duplication.
- **camelCase vs the registry key.** The clone produced `mrsMop.md`, `mrsMop.json` and a `mrsMop.*` i18n namespace while the registry key is `mrsmop`. Manuals, both locale sets and every namespace reference are renamed — those files are keyed by the registry name, so the guards caught it.
- **Persistence tests no longer build history with `Deal`.** They construct an explicit two-card board and move, so they don't depend on the shuffle; a deal-dependent fixture here fails only in the full-package run, which is how I burned a cycle on #6192.

## Test plan

- [x] `go test -tags test ./internal/...`
- [x] `golangci-lint run ./internal/...` — 0 issues
- [x] `goimports -l` on every changed Go file — clean
- [x] `cd frontend && bun run check` — passes at 329 games
- [x] `cd frontend && bun run typecheck`
- [x] 222 targeted frontend tests
- [x] `count-audit.sh` — ✅ all count assertions consistent (329 games)
- [x] `GOOS=js GOARCH=wasm go build -tags extra4` — no stranded symbols
- [x] OpenAPI CRLF verified intact (61,964 CRLF, 0 bare LF)
- [x] Source-game sweep across `Spider`/`spider`/`スパイダー` — only deliberate contrast notes remain
- [ ] Full frontend suite + E2E — left to CI


yuta-yoshinaga:
All four fixed in 534b2e48b — every one was real, and finding 1 is the one I most needed pointed out.

**1. `UnmarshalJSON` defaulting to 1-suit.** You're right that this contradicts the PR's own stated principle, and the reason it happened is worth naming: I unified the default across `ResetWithConfig`, `DefaultMrsMopConfig` and `NewDefaultMrsMop` and stopped there — three of four entry points. The one I missed is the only one that runs on a *restore*, so a saved or corrupt KV game came back as the relaxed variant while every fresh game was correct. Now an explicit switch matching the other three, with a test over `{}`, `{"df":0}` and `{"df":99}` plus the two variants that must survive round-trip.

**2. 10 vs 13 columns.** Fixed, and your diagnosis of *why* the 222 tests missed it is the part I want to record: `makeTableau` in the page test also padded to 10, so **the mock agreed with the bug rather than with the backend contract**. A test whose fixture is cloned alongside the code can only confirm the clone. Both the page and the test now derive from a single `MRSMOP_COLUMN_COUNT = 13` with a sync comment to `domain.MrsMopTableauCnt`.

**3. Stock-pile tutorial step.** Removed, with its i18n key. A spotlight on an element the page deliberately never renders.

**4. Dead deal-from-stock hint.** Removed — branch, i18n string, and the test that fabricated `stockCount: 10` to reach a state the real API cannot produce. While there I found the CLI formatter printing a permanent `stock: 0` for the same reason; that's gone too, and the remaining `stockCount: 50` fixtures across four test files are now `0`, matching what the domain actually returns.

On your two notes: agreed on `FaceUp`/`autoFlipTableau` — it's the shared `TableauCard` shape, so keeping it costs nothing and diverging would be worse. The `m <fromCol> <toCol>` shorthand always erroring is, as you say, pre-existing across Spider and 27 other cloned games; I've recorded it rather than widening this PR, since fixing it properly is a cross-cutting change to the shared CUI move parser.

Gates re-run: `go test -tags test ./internal/...`, `golangci-lint` 0 issues, `goimports` clean, `bun run check`, `typecheck`, 94 targeted frontend tests.

PR #6201, opened 2026-08-21
feat(followthequeen): add Follow the Queen game

yuta-yoshinaga:
Closes #5434

Seven Card Stud with a **moving wild rank**: a face-up Queen makes the rank of the very next face-up card wild for everyone, and the four Queens are wild too. A later face-up Queen cancels the previous wild and arms a new one, so the best hand at the table can change on a single card. If that Queen is the last face-up card of the deal, only the Queens stay wild.

## Why the evaluator is new

The wilds are ordinary cards, not jokers, so `evalFiveCardHandWithJokers` does not apply — it keys on `IsJoker` and caps at two, while this game can have eight live wilds (4 Queens + 4 of one rank). `evalFiveCardHandWithWilds` substitutes each wild by exhaustive search and keeps the best result, which makes **five of a kind** reachable and the strongest hand in the game.

One branch in it (`len(wildIdx) >= 4 → five of a kind`) is a **performance shortcut that cannot change the answer** — the brute force reaches the same conclusion after 52⁴ combinations. It is documented as an optimisation rather than claimed as tested, because no test can observe it.

## Making the rule visible

A wild nobody can see is a hand nobody can count, so `wildRank` is surfaced on every output surface:

| Surface | What it shows |
|---|---|
| Web API | `wildRank` on the response (`0` = none yet), documented in `openapi.yaml` |
| Page header | `Wild: 7` / `Wild: none yet`, face ranks as letters |
| CUI | a `ワイルド:` line above the pot, on every board |
| Both hints | one wild = a pair at worst, two = trips or better |

The hint change matters: plain-stud advice misreads the same three cards by two steps once a wild is live. The CUI hint asks the **domain's** `IsWild` rather than re-deriving "Queens are always wild" in the presenter.

Player-side wild state is stored as a value, not a predicate — the Worker is stateless per request and a closure would not survive the KV round trip.

## Bucket

`casino`, not the `extra` the issue suggested. The shared betting primitives it depends on (`HoldemPlayStyle`, `BettingLimitType`, `SidePot`, `BettingHumanProfile`) are casino-tagged; retagging elsewhere strands them. Verified with `GOOS=js GOARCH=wasm go build -tags <w>` for all seven workers.

## Clone cleanup

Cloning `sevencardstud` dragged in two modes this game cannot enter (`NewDefaultFollowTheQueen` leaves both flags false):

- **Hi-Lo** — with up to eight wilds every low is a wheel, so the split is meaningless here.
- **Razz** — same reason, inverted.

[...]

PR #6196, opened 2026-08-21
feat(rankandfile): add Rank and File solitaire game

yuta-yoshinaga:
## Summary

Adds **Rank and File** (`rankandfile`, bucket `extra4`) — the Forty Thieves variant whose ten columns are dealt four deep with the bottom three face down, whose tableau builds down by **alternating colour** rather than by suit, and where any descending run moves as a unit.

Cloned from `fortythieves` and diverged at exactly three points. Each divergence is pinned by a test that carries a **negative control the source game would fail**, so the tests cannot pass against a leftover Forty Thieves predicate:

| Divergence | Where | Negative control |
|---|---|---|
| Alternating colour, not same suit | `canPlaceOnTableau` | `reject the same colour` — ♠5 on ♠6 is legal in Forty Thieves, rejected here |
| Any descending run moves as a unit | `MoveTableauToTableau` | a broken run is refused; a partial run below the top is accepted |
| Three cards face down per column | deal / `AllFaceUp` / `autoFlipTop` | `AllFaceUp` no longer just reports `len(stock) == 0`; a face-down card is not a legal destination or source |

## Also fixed

The CUI shorthand `m <from> <to>` — the controller documents it as "tableau-to-tableau top card" but passed `cardIndex = -1`, which the range check rejected outright, so the shorthand printed in `help` always failed. `-1` now resolves to the column's top card; **other** negative indices still error (negative control included).

## Deliberate divergence from the issue

#5450 specifies a 96-card deal across columns of nine or ten with the bottom three cards **face up**. That is not Rank and File as catalogued — both Parlett and PySolFC give 10 columns x 4 cards, three down and one up, with the remaining 64 cards as stock. The implementation follows the game; flagging it here rather than silently.

## Verification

- \`go test -tags test ./internal/...\` — pass
- \`golangci-lint run ./internal/...\` — 0 issues
- \`goimports -l\` on every touched Go file — 0
- \`GOOS=js GOARCH=wasm go build -tags extra4 ./cmd/workers/extra4\` — builds (no stranded symbol)
- \`count-audit.sh\` — all ✅, 330 games
- frontend \`bun run check\` / \`typecheck\` / targeted vitest — pass
- \`api/openapi.yaml\` still CRLF-only and parses as YAML
- The frontend face-down tests and the hint-live-region registration were both mutation-verified (deleting \`rf-hint-announcement\` still fails the guard by name)

Closes #5450

PR #6202, opened 2026-08-22
feat(ramsch): add Ramsch game

### LINKED ISSUE #5453: Ramsch (ラムシュ) の追加
## 概要 (Overview)
ラムシュ (Ramsch) はドイツのスカート地方に由来する3人用トリックテイキングゲーム。本来は Skat で誰も入札しなかった際の副次ルールだが、単独の「逆取り」ゲームとして広く遊ばれている。得点を「取らない」ことが目標で、最も多く点を取ったプレイヤーが失点する点が最大の特徴。

- **メカニクス**: 既存の `skat` はビッド制で切り札スートを宣言し高得点を競うが、Ramsch にはビッドが存在せず、常に4枚のジャックのみが固定の切り札となる逆転ゲーム。得点の最大化ではなく最小化を争う点で `skat` と対照的。
- **地域カバレッジ**: ドイツ語圏スカート系ゲームのバリエーションを拡充し、逆取り (misère) 系ドイツ実装の第一号となる。
- **実装の見通し**: `internal/domain/Skat.go` の32枚デッキ生成・トリック処理・カード点数計算 (A=11, 10=10, K=4, Q=3, J=2) をそのまま流用できる。新規に書くのは、ビッドフェーズを丸ごと省いた配札直後開始のフロー、ジャック固定切り札判定、得点最小化 CPU 戦略、Durchmarsch (全13トリック総取りで逆転勝利) 判定のみ。

## 仕様案 (Requirements)
- **カテゴリ（バケット）**: `extra2` を推奨。ただし Category は Cloudflare Worker のサイズバケットであり、**2026-08-16 実測で6 worker の残容量は合計 220 KB しかない**（solo 51.3KB / extra3 47.5KB / extra2 46.3KB / classic 32.5KB / casino 23.2KB / extra 19.1KB）。本バッチは 7 番目の worker 新設 #5421 が前提であり、着手時に `.claude/skills/rebucket-game/scripts/measure.sh` で実測してバケットを確定すること。
- **人数**: 3人 (1人プレイヤー vs CPU 2人)
- **デッキ**: 32枚 (スカート牌)
- **主要ルール**:
  1. 各プレイヤーに10枚ずつ配り、残り2枚をスカット(伏せ札)に置く。ビッドは行わない。
  2. 4枚のジャック (クラブ>スペード>ハート>ダイヤ) が常に最強の切り札。残りは各スートのAが最強。
  3. リードスート必従で13トリックを行い、各トリックの獲得者は次のリードを取る。
  4. 各カードの点数 (A=11,10=10,K=4,Q=3,J=2、合計120点) を集計し、獲得点数が最も多いプレイヤーがその点数分を失点する。
  5. 誰か1人が全13トリックを総取りした場合 (Durchmarsch) は、その他2人が敗北として得点を負う逆転勝利。
  6. 複数ラウンドを繰り返し、累計失点が最も少ないプレイヤーが勝者。

- 再利用: `internal/domain/Skat.go` の32枚デッキ生成・トリック判定・スート必従ロジックを流用し、新規に書くのはビッド省略フロー、ジャック固定切り札、得点最小化 CPU 戦略、Durchmarsch 判定のみ。
- 新規ドメイン型: `internal/domain/Ramsch.go`, `RamschConfig.go`, `RamschPlayer.go`

---
_新規ゲーム候補バッチ（2026-08-16, 49件）により起票。実装時は docs/new-game-checklist.md に従い、バケットは着手時に .claude/skills/rebucket-game/scripts/measure.sh で実測すること。ブロッカー: #5421（6 worker の gzip 残容量が合計 220 KB しかなく、7番目の worker 新設が前提）（スラッグ案: `ramsch`）_

yuta-yoshinaga:
Closes #5453

Skat turned inside out. **There is no auction** — nobody bids, nobody picks up the skat, nobody declares a contract — so play starts the moment the cards are dealt, and the **four jacks are always the trumps** (♣J > ♠J > ♥J > ♦J). A jack never counts as a card of its printed suit.

**Card points are penalty points.** Whoever collects the most of the 120 loses exactly that many; a tie for most penalises everyone tied, since there is no basis for singling one seat out. **The two face-down skat cards go to the winner of the last trick**, so all 120 always land on somebody rather than a few vanishing. The one way to win outright is **Durchmarsch**: sweep all ten tricks and the other two lose 120 each.

## What cloning cost

`skat` was the right source for the trick engine, but the clone arrived carrying an entire auction. Removing it was most of the work:

| Removed | Why |
|---|---|
| Bid ladder, `PlayerBid`/`CpuBid`, 4 auction phases | No auction exists |
| Skat pickup + discard | Nobody touches the skat |
| Game declaration, `RamschGameType` (Suit/Grand/Null) | Trump is fixed |
| Matadors, base values, Schneider/Schwarz/Hand bonuses, `ScoreBreakdown` | A contract scoring model this game has no contract for |
| `skatBidEstimate`, `skatScoreBonuses` frontend utils | Dead with the auction |

Leaving any of it would have shipped unreachable phases and a scoring model the game does not have — the same lesson as Follow the Queen's Hi-Lo in #6201.

## The CPU had to be inverted, not tuned

Skat's "win the trick if you can" is exactly wrong here. The rewrite:

- **ducks whenever it can**, and
- **ducks with its _highest_ losing card**, so aces do not survive into the endgame, and
- takes a trick **as cheaply as possible** only when it cannot avoid one.

## `Ramsch_json.go` is new and load-bearing

Every field on `Ramsch` is unexported, so plain `json.Marshal` returns `{}` — no error — and the Worker would restart the board on every request while the local CLI never noticed.

**Skat itself still ships this way.** Measured: `json.Marshal(NewDefaultSkat())` after `Reset()` returns **2 bytes, `{}`**. Every registered KV game with only unexported fields and no marshaller has the same problem. Out of scope here, but it should have an issue.

## The issue's spec was wrong about trick count

It said 13 tricks. 32 cards ÷ 3 players = 10 each plus a 2-card skat, so it is **ten** — `SkatTricksPerRound = 10` already said so. Recorded in `docs/games.md`.

## Mutations verified

| Mutation | Result |
|---|---|
| scoring not inverted (most points wins) | FAIL ✅ |
| Durchmarsch not reversed | FAIL ✅ |
| Durchmarsch *detection* disabled | FAIL ✅ |
| Durchmarsch off-by-one (9 tricks counts) | FAIL ✅ |
| skat not awarded to the last trick | FAIL ✅ |
| ties penalise only the first seat | FAIL ✅ |
| ten ranks below king | FAIL ✅ |
| double-resolve guard removed | FAIL ✅ |
| CPU wins instead of ducking | FAIL ✅ |
| drop `Skat` / `CurrentTrick` / `Config` from the wire | FAIL ✅ ×3 |
| whole JSON back to `{}` | FAIL ✅ |
| page: loser badge on every seat / skat always shown / no Durchmarsch badge / play not gated | FAIL ✅ ×4 |

Two branches survive mutation and are **documented as unobservable rather than claimed as tested**: `effectiveSuit`'s jack case (every caller checks `isTrump` first) and `TrumpCards` in the wire format (all 32 cards are dealt at reset, so the deck is never drawn from again).

## Three deal-dependent tests fixed

`go test ./internal/domain` failed **2 runs in 4** on this branch and **0 in 3** on develop — the new tests shift the shuffle sequence and expose latent deal-dependence that was already there. None is caused by Ramsch:

- **`RankAndFile`** asserted card index `-1` is invalid, but `-1` is the CUI shorthand for "the column's top card" that #6196 added — a legal move on the right deal. Now uses `-2`, plus a test pinning the shorthand so re-breaking it is caught.
- **`Stalactites`**'s "empty cell" case never emptied a cell — this game deals *into* them — so it passed because the card could not be placed. Now it clears the cell, with a negative control that the move succeeds once it can.
- **My own `FollowTheQueen`** all-in test read absolute card counts, but `Reset()` runs CPU turns and can pass third street. Now measures the increment.

`./internal/domain` now runs clean 3/3.

## Also

Card points were tracked twice — on the round and on the player — and the web presenter read the stale copy. Collapsed to one source.

## Test plan

- [x] `go test -tags test ./...`
- [x] `golangci-lint run ./...` — 0 issues
- [x] all 7 worker targets build for WASM
- [x] `bun run typecheck` / `check`; Vitest page suite
- [x] `count-audit.sh` — all ✅ (336 games)
- [x] `./internal/domain` run 3× clean; usecase suite 60× clean
- [ ] CI


yuta-yoshinaga:
## All four findings fixed

Thanks — the first one was serious and my tests were blind to it.

### 🔴 `checkGameEnd` was unsatisfiable

Confirmed exactly as described. Ramsch cumulative scores are penalty points and are always ≤ 0; `TargetScore` is bounded ≥ 1 on **every** path (`Validate`, `RamschWebConfig.ToConfig`, the CUI `settarget`, and the frontend's 100/250/500/1000 options). `cumulativeScore >= TargetScore` can never hold, so `RamschPhaseGameEnd` and the whole frontend win state were unreachable — the match just ran forever.

Fixed to compare against `-TargetScore`. I also changed the log line: crossing a penalty limit does not make you the winner, it ends the match with you as the one everyone beat. The doc on `TargetScore` says so now too.

**The reason nothing caught this is the important part**: no test in any of the four files touched `checkGameEnd` or `GetGameEndFlag`. There is one now, and three mutations fail against it:

| Mutation | Result |
|---|---|
| restore `cumulativeScore >= TargetScore` (the original) | FAIL ✅ |
| off-by-one — fires a round early | FAIL ✅ |
| never ends | FAIL ✅ |

### 🟡 The other three were one leftover with three faces

You are right that the PR description claimed the auction was gone while its **input surface** was still there. I removed the dispatch paths and stopped.

- `RamschWebInput` — dropped `accept`/`pickup`/`discardA`/`discardB`/`gameType`/`trumpSuit`.
- `ramschApi.exec` — command union trimmed to `reset | play | next | nextround | hint | log` and args to `cardIndex`/`config`. Your point that it type-checked and then silently hit the unknown-command fallback is exactly the failure mode; it also disagreed with the `RamschRequest` OpenAPI schema, which was already correct.
- `RamschCuiController` doc comment — trimmed to the commands the registered list actually contains.

Each now has a comment saying *why* the field is absent, so the next clone does not re-add it.

### Re-verified

`go test -tags test ./...` (domain package 2× clean), `golangci-lint run ./...` 0 issues, all 7 worker targets, `bun run typecheck` / `check`, Vitest page suite.

PR #6203, opened 2026-08-22
feat(seventwentyseven): add Seven Twenty-Seven game

### LINKED ISSUE #5433: Seven Twenty-Seven (セブン・トゥエンティセブン) の追加
## 概要 (Overview)

Seven Twenty-Seven はアメリカの「dealer's choice」ホームゲームの定番の一つで、7に近い手と27に近い手の二方向でポットを二分する。絵札は0.5、エースは1か11を選べるという独自のカード価値が、単純な高低勝負に読み合いを持ち込む。

- **メカニクス**: 既存の `guts`（アンティ→カード公開→続けるか降りるかを繰り返すポーカーバリアント）に対し、Seven Twenty-Seven は続けるか降りるかの判断に加え、追加カードを引いて7側・27側どちらの持ち点に寄せるかという二方向の駆け引きが乗る点が異なる。
- **地域カバレッジ**: アメリカの dealer's choice 系ホームゲームの系統を補う。
- **実装の見通し**: `internal/domain/Guts.go` のアンティ・繰り返しベッティング・降り判定のループを流用できる。新規に書くのは、絵札=0.5・エース=1 or 11というカード価値評価器と、7に一番近い手／27に一番近い手（超過は失格）の二方向判定、およびポットの二分配分ロジック。

## 仕様案 (Requirements)

- **カテゴリ（バケット）**: `casino` を推奨。ただし Category は Cloudflare Worker のサイズバケットであり、**2026-08-16 実測で6 worker の残容量は合計 220 KB しかない**（solo 51.3KB / extra3 47.5KB / extra2 46.3KB / classic 32.5KB / casino 23.2KB / extra 19.1KB）。本バッチは 7 番目の worker 新設 #5421 が前提であり、着手時に `.claude/skills/rebucket-game/scripts/measure.sh` で実測してバケットを確定すること。
- **人数**: 2〜7人（1人vsCPU複数卓として実装）
- **デッキ**: 52枚
- **主要ルール**:
  1. 全員がアンティを払い、数枚ずつ配る（絵札=0.5点、エース=1点または11点をプレイヤーが選択、それ以外は数字通り）。
  2. 各ラウンドで追加カードを引くか降りるかを選ぶベッティングを繰り返す。
  3. 合計点が7を超えた、または27を超えたプレイヤーはその側の勝負から失格する。
  4. 最終的に7に最も近い（超えていない）プレイヤーと27に最も近い（超えていない）プレイヤーを決定する。
  5. ポットを7側の勝者と27側の勝者で二分する（同一プレイヤーが両方を制した場合は総取り）。

- 再利用: `internal/domain/Guts.go` のアンティ・反復ベッティング・降り判定ループを流用し、新規に書くのは絵札=0.5/エース選択のカード価値評価器、7側/27側の二方向判定、ポット二分配分のみ。
- 新規ドメイン型: `internal/domain/SevenTwentySeven.go`, `SevenTwentySevenConfig.go`, `SevenTwentySevenPlayer.go`

---
_新規ゲーム候補バッチ（2026-08-16, 49件）により起票。実装時は docs/new-game-checklist.md に従い、バケットは着手時に .claude/skills/rebucket-game/scripts/measure.sh で実測すること。ブロッカー: #5421（6 worker の gzip 残容量が合計 220 KB しかなく、7番目の worker 新設が前提）（スラッグ案: `seventwentyseven`）_

yuta-yoshinaga:
Closes #5433

The pot splits **two ways**: the hand closest to **7** without going over and the hand closest to **27** without going over each take half, and one player taking both **scoops** the lot.

## Card values are the whole point

| Card | Points |
|---|---|
| A | **1 or 11**, chosen independently per ace |
| 2–10 | pip value |
| J / Q / K | **0.5** |

A-A-5 makes **exactly 7** and **exactly 27** — the hand you play for.

**Everything is held as integers scaled by ×2.** A face card is 1, a number card is 2n, and the targets are 14 and 54. The half-point rule stays exact with no float comparison anywhere; only the display divides. Same choice Call Break makes with its ×10 scores.

## Cloning `guts` gave less than it looks

Guts supplied the ante, pot and carry-over — but its shape of play is wrong here. Guts is **one simultaneous in/out declaration** with exactly two phases; this game repeats "card or stand" until everybody has stood. The draw loop, the two-way settlement and the CPU are all new, and the poker-hand evaluation, the declare guide and its golden-vector fixture went out.

## The CPU had to pick a side, not be tuned

Chasing 27 unconditionally made **the low half of the game disappear** — measured at **0 low-side survivors across 12 seats**. Any hand alive for 7 is one card from death, so a CPU that always draws never leaves one. It now compares how far each target is and either protects a low hand or draws toward 27.

After the fix, over 200 rounds: **low side had a winner 135 times, high side 200, and a player scooped 4 times.** Both halves are live.

## Standing is final, and the round finishes itself

Waiting for the human after they stood would hang the board — they have no legal move left, so nothing could advance it. `TakeCard` keeps playing CPU passes until everybody has stood, then settles.

## Mutations verified

| Mutation | Result |
|---|---|
| face card worth 1 point instead of 0.5 | FAIL ✅ |
| ace stuck at 1 | FAIL ✅ |
| bust not detected | FAIL ✅ |
| picks the lowest total under the target | FAIL ✅ |
| high target unscaled (27 instead of 54) | FAIL ✅ |
| standing players still get cards | FAIL ✅ |
| pot not split — low side takes all | FAIL ✅ |
| an empty side forfeits its half | FAIL ✅ |
| CPU always chases 27 | FAIL ✅ |
| the round does not finish itself | FAIL ✅ |
| page: actions after standing / scoop as two halves / one score only / no carry panel | FAIL ✅ ×4 |

## Surfaces

The two targets and the unusual card values are stated on **every** board in both interfaces — this is the rule players arrive without. Each seat shows `low / high`, with `-` for a side that has gone over, and the hint names **which side it is chasing** (`chase_seven` vs `chase_twentyseven`); "draw" alone is not advice when the two targets pull opposite ways.

Opponents' scores stay hidden until the round settles — the totals *are* the hand, so hiding one means hiding both.

## Test plan

- [x] `go test -tags test ./...` (domain package 2× clean)
- [x] `golangci-lint run ./...` — 0 issues
- [x] all 7 worker targets build for WASM
- [x] `bun run typecheck` / `check`; Vitest page suite (9 tests)
- [x] `count-audit.sh` — all ✅ (337 games)
- [ ] CI


yuta-yoshinaga:
## Patch coverage fixed (69% → per-file 85–100%)

The two biggest holes were self-inflicted: I rewrote the CLI parser and formatter for this game and **deleted the cloned tests without replacing them**, leaving them at **8%** and **3%**.

| File | Now |
|---|---|
| `seventwentyseven_score.go` | 100% |
| `SevenTwentySevenConfig.go` | 100% |
| `SevenTwentySevenPlayer.go` | 99% |
| `SevenTwentySevenInteractor.go` | 97% |
| `SevenTwentySevenCuiPresenter.go` | 92% |
| `SevenTwentySevenWebPresenter.go` | 88% |
| `SevenTwentySeven.go` | 85% |
| `SevenTwentySevenCuiController.go` / `WebController.go` | 100% |

The new tests pin what actually changed rather than padding the number:

- **The parser** — `card`/`stand` replace Guts's `in`/`out`, and the old commands are **rejected**, not silently accepted.
- **The help table** — checked against the parser, but on the right property: that it never advertises a command the parser does not *know*. My first attempt appended a fixed argument to every documented token and failed on `sc`, whose minimum is 10 — that was the test being wrong, not the help, so it now distinguishes a range error from an unknown command and carries a negative control.
- **Config validation on both sides of every bound**, including starting chips below the ante, which would deal a table that cannot play a single round.
- **The player's JSON round trip** and its rejection of negative chips.
- **The match-end path** and the carried pot seeding the next round's ante.

Re-verified: `golangci-lint` 0 issues, `./internal/domain` clean, `bun run typecheck` / `check`, 63 frontend tests across the three suites.

yuta-yoshinaga:
## Review findings fixed — plus one the review didn't reach

The automated reviewer truncated at 3 of 6 items again, so I ran the remaining layers myself. Between the two, six real defects — two of them things a user would see immediately.

### 🔴 The hint tooltip rendered a raw identifier

`reason` was built as `` `seventwentyseven:${hint.reason}` ``. i18next's default `nsSeparator` is `:`, so that resolves **namespace** `seventwentyseven`, **key** `chase_seven` — which does not exist, because I wrote the locale keys as flat `hintReasonChaseSeven`. Turning hints on printed `chase_seven` verbatim, and all six reason strings were dead.

Now `` `hint.${reason}` `` with a nested `hint` object keyed by the snake_case reason the backend emits — the Guts convention, which is what the rest of the repo uses.

### 🔴 The scoop banner always claimed a pot of 0

`finishRound` zeroes the pot *before* flipping to the result phase, so `state.pot` is always 0 there. The one moment this game is built around announced the prize as **zero**.

The message no longer names an amount — matching the CUI string, which never did. **The existing test missed it because its fixture carried `pot: 40`, a value the server cannot return at that phase.** Fixture corrected; the assertion now fails on either a literal `0` or an unfilled placeholder, verified by reinstating `{{pot}}`.

### 🟡 `hintRequested` leaked `{{hint}}`

The presenter sets `MessageCode` but never `MessageParams`, so the token rendered raw. It was the only `{{hint}}` in the whole of `common.json`; every other game uses a parameterless string.

### 🟡 OpenAPI documented bounds the server does not honour

`ante`, `startingChips` and `targetRounds` had minima but **no maxima**, while `ApplyBoundedInt` silently falls back to the default outside the range. A client following the spec posts `ante: 5000`, gets a `200`, and is quietly playing at `10`. Maxima documented; `next` added to the command enum since the dispatcher accepts it.

### One the review didn't reach: `carryCount` was in neither direction of the wire format

Found by auditing the struct field list against the marshaller — the same check that caught Big Ben's stock and Follow the Queen's wild rank. A Worker session that carried the pot twice came back reporting **"carry #0"**. Both directions now, with a test that carries twice before serialising so it cannot pass on `0 == 0`. Dropping either direction fails it.

### Also

- The **final tutorial step could never anchor**: `dataTutorial=` is a different attribute spelling from the `data-tutorial=` my rename swept, so it kept the pre-rename value. Every step now has an anchor and every anchor a step (the players div had one without the other).
- **Stale Guts prose** survived in the package doc, the CUI controller's command list and the player's field comments — the package doc still described simultaneous in/out declarations and pair comparisons.
- **`docs/cloudflare-workers.md`'s extra4 row was missing five games** — `fourteenout`, `mrsmop`, `narcotic`, `perseverance`, `rankandfile`. All five are mine from earlier in this batch, so the drift is mine. The row is regenerated from the registry, and I checked all seven rows: every one now matches.

### Re-verified

`golangci-lint` 0 issues, full Go suite, `bun run typecheck` / `check`, 63 frontend tests across the three suites.

PR #6204, opened 2026-08-22
fix(ramsch): resolve the hint reason instead of printing its identifier

yuta-yoshinaga:
Ramsch's hint tooltip has been printing `avoid_points` at the user since #6202 merged.

## What happened

The hint factory built its reason as `` `ramsch:${reason}` ``. i18next's default `nsSeparator` is `:`, so that is parsed as **namespace `ramsch`, key `avoid_points`** — and no such key exists, because the locale file spells them `hintReasonAvoidPoints`.

| | |
|---|---|
| Go emits | `avoid_points`, `lead_low`, `forced_discard` |
| Frontend asked for | `ramsch:avoid_points` → ns `ramsch`, key `avoid_points` |
| Locale file had | `hintReasonAvoidPoints`, `hintReasonLeadLow`, `hintReasonForcedDiscard` |

So the tooltip showed the raw identifier and all three reason strings were dead.

## How it was found

While fixing the **identical** mistake in Seven Twenty-Seven (#6203) — I made it twice in a row. A sweep of every hint factory shows those two were the only ones using the `ns:key` form; the other ~120 call sites use `hint.${...}` or `hintReason.${...}`.

## The fix

`` `hint.${reason}` `` with a nested `hint` object keyed by the snake_case reason the backend actually emits, matching Guts and the rest of the repo.

## Why the test is shaped the way it is

**Asserting that a string comes back passes against this bug** — a raw identifier is a perfectly good string. The test asserts the reason **resolves to real text**:

```ts
const text = i18n.t(hint.reason, { ns: 'ramsch' });
expect(text).not.toBe(key);       // i18next returns the key on a miss
expect(text).not.toContain(reason); // and the raw identifier is not the answer
```

with a **negative control** showing an unknown reason still comes back unresolved, so the assertion is not vacuous.

Reinstating the `ns:key` form fails **4 of its 6 cases**.

## Test plan

- [x] `bun run typecheck` / `bun run check`
- [x] Vitest — 6 tests, mutation-verified against the shipped bug
- [ ] CI

### DIFF
--- a/frontend/src/i18n/locales/en/ramsch.json
+++ b/frontend/src/i18n/locales/en/ramsch.json
@@ -34,7 +34,9 @@
   "skatLabel": "Skat (goes to whoever wins the last trick)",
   "loserBadge": "most points — loses",
   "durchmarschBadge": "Durchmarsch",
-  "hintReasonAvoidPoints": "duck with a card worth nothing and take no points",
-  "hintReasonLeadLow": "lead something cheap so you do not put points on the table",
-  "hintReasonForcedDiscard": "only point cards are legal here, so give up the cheapest"
+  "hint": {
+    "avoid_points": "duck with a card worth nothing and take no points",
+    "lead_low": "lead something cheap so you do not put points on the table",
+    "forced_discard": "only point cards are legal here, so give up the cheapest"
+  }
 }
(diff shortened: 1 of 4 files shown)


yuta-yoshinaga:
Patch coverage flagged the `|| 'avoid_points'` fallback as the one uncovered line.

It **is** reachable: the guards above check `hint` and `cardIndex` but not `reason`, so a response with a card index and an empty reason lands there. The server always sets one, but a version skew or a corrupted KV value would not — and without the fallback an empty reason builds the key `hint.`, which resolves to nothing and puts an empty tooltip on screen. So it is covered rather than removed.

`ramschHint.ts` is now **100%** on statements, branches, functions and lines.

